PR #4, opened 2015-05-07
Using expression-bodies methods

hishamco:


Eilon:
@DamianEdwards thoughts? This is a syntax-only change, so it's up to you.

DamianEdwards:
Yeah this is ok. I used them elsewhere in this repo but wasn't consistent when I added more stuff.

hishamco:
I used them whenever the method body is single line, further more we are trying to make the code simple and clean using this feature

PR #11, opened 2015-05-13
Move CultureInfoCache into its own package.

DamianEdwards:
Moves `CultureInfoCache` into a new package `Microsoft.Framework.Globalization.CultureInfoCache`

@davidfowl @Eilon @kirthik @muratg


DamianEdwards:
Feeling a little lonely here... @davidfowl @lodejard @Eilon @kirthik @muratg

src/Microsoft.Framework.Globalization.CultureInfoCache/project.json
```diff
@@ -0,0 +1,19 @@
...
+  },
+
+  "frameworks": {
+    "dnx451": { },
+    "dnxcore50": {
+      "dependencies": {
+        "System.Collections": "4.0.10-*",
+        "System.Collections.Concurrent": "4.0.10-*",
+        "System.Globalization": "4.0.10-*",
+        "Microsoft.CSharp": "4.0.0-*"
```
Eilon: These are sooooo not sorted!

Eilon:
:shipit: 

We should talk more to the .NET folks about this... ultimately nothing will remain in `Microsoft.Framework.*` so we need to think whether this goes to .NET, or if it goes somewhere else (`Microsoft.Dnx.*`). Doesn't have to change now, but need to think about this.

davidfowl:
We should stay away from Microsoft.Dnx for anything that's not in the dnx repository.

davidfowl:
:shipit:

Issue #15, opened 2015-05-15
ResourceManagerStringLocalizer should cache GetEnumerator result


DamianEdwards:
Every call to `ResourceManagerStringLocalizer.GetEnumerator` walks the specified `CultureInfo` hierarchy to build out a full set of resource names, before calling `ResourceManager.GetString` for every resource name. Given that resources are only loaded once by `ResourceManager` we should cache at least the culture walk, and probably the entire enumeration result, per culture.

PR #20, opened 2015-05-19
Cache the walking of the CultureInfo tree in ResourceManagerStringLocalizer

DamianEdwards:
Implemented caching of the walk of the `CultureInfo` tree to get resource names. Includes tests.

#15 

@davidfowl @Eilon @muratg

### DIFF
--- a/src/Microsoft.Framework.Localization/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.Framework.Localization/Properties/AssemblyInfo.cs
@@ -1,0 +1,6 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.Framework.Localization.Test")]
--- a/src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
+++ b/src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
@@ -18,9 +18,11 @@ namespace Microsoft.Framework.Localization
     /// </summary>
     public class ResourceManagerStringLocalizer : IStringLocalizer
     {
-        private readonly ConcurrentDictionary<MissingManifestCacheKey, object> _missingManifestCache =
-            new ConcurrentDictionary<MissingManifestCacheKey, object>();
-        private readonly ConcurrentDictionary<string, IList<string>> _cache = new ConcurrentDictionary<string, IList<string>>();
+        private readonly ConcurrentDictionary<string, object> _missingManifestCache =
+            new ConcurrentDictionary<string, object>();
+
+        private readonly ConcurrentDictionary<string, IList<string>> _resourceNamesCache =
+            new ConcurrentDictionary<string, IList<string>>();
 
 
         /// <summary>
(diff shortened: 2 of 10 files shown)


src/Microsoft.Framework.Localization/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.Framework.Localization.Test")]
```
muratg: Not sure what @davidfowl thinks, but in some of the projects, we moved Internal classes to an .Internal namespace and made them public for testability instead. We don't seem to be following this on all projects, so just something to think about.

src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -185,43 +184,34 @@ private IEnumerable<string> GetResourceNamesFromCultureHierarchy(CultureInfo sta
...
-            {
-                var other = obj as MissingManifestCacheKey;
-
-                if (other != null)
+                var names = new List<string>();
+                using (var cultureResourceStream = ResourceAssembly.GetManifestResourceStream(key))
+                using (var resources = new ResourceReader(cultureResourceStream))
                 {
-                    return Equals(other);
+                    foreach (DictionaryEntry entry in resources)
```
muratg: nit: var

test/Microsoft.Framework.Localization.Test/ResourceManagerStringLocalizerTest.cs
```diff
@@ -0,0 +1,139 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Resources;
+using Moq;
```
muratg: Does this work on CoreCLR now?

test/Microsoft.Framework.Localization.Test/project.json
```diff
@@ -0,0 +1,16 @@
...
+        "xunit.runner.dnx": "2.1.0-*",
+        "Microsoft.Framework.Localization": "1.0.0-*"
+    },
+
+    "commands": {
+        "test": "xunit.runner.dnx"
+    },
+
+    "frameworks": {
+        "dnx451": { }
```
muratg: No CoreCLR?

muratg:
Looks good, :shipit:

src/Microsoft.Framework.Localization/Properties/AssemblyInfo.cs
DamianEdwards: In this case it's just an internal method to clear a global static cache, not types.

src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
DamianEdwards: `var` doesn't work here because it's not `IEnumerable<T>`, just plain old `IEnumerable`. Yay old APIs! :smile:

test/Microsoft.Framework.Localization.Test/ResourceManagerStringLocalizerTest.cs
DamianEdwards: Not that I can see. At least not the version on nuget.org

test/Microsoft.Framework.Localization.Test/project.json
DamianEdwards: Because it's using Moq.

test/Microsoft.Framework.Localization.Test/ResourceManagerStringLocalizerTest.cs
DamianEdwards: See https://github.com/Moq/moq4/issues/168 and https://github.com/castleproject/Core/issues/90

Still coming

hishamco:
@DamianEdwards for better organization, can we introduce an interface named ```IResourceCache``` that resource cache adapters have to implement

DamianEdwards:
@hishamco This isn't something that should be abstracted. It's an implementation detail of this specific `IStringLocalizer`. It doesn't make sense to create a provider/adapter model here.

DamianEdwards:
Merged in 9384848cc7c0a4faf92c3d2157fcd7621ac3fe57

Issue #18, opened 2015-05-19
IStringLocalizer should have a method to get string list rather than implement IEnumerable itself


DamianEdwards:
'Cos it's just weird right now.

PR #22, opened 2015-05-20
Add IStringLocalizer.GetAllStrings method

DamianEdwards:
@davidfowl @muratg 

#18 #23

### DIFF
--- a/src/Microsoft.Framework.Localization.Abstractions/StringLocalizerExtensions.cs
+++ b/src/Microsoft.Framework.Localization.Abstractions/StringLocalizerExtensions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved. 
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
+using System.Collections.Generic;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.Framework.Localization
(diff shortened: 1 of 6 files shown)


src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -94,6 +94,28 @@ public IStringLocalizer WithCulture(CultureInfo culture)
...
+        /// <param name="culture">The <see cref="CultureInfo"/> to get strings for.</param>
+        /// <returns>The strings.</returns>
+        protected IEnumerable<LocalizedString> GetAllStrings([NotNull] CultureInfo culture)
+        {
+            var resourceNames = GetResourceNamesFromCultureHierarchy(culture);
+
+            foreach (var name in resourceNames)
+            {
+                var value = GetStringSafely(name, culture);
+                yield return new LocalizedString(name, value ?? name, resourceNotFound: value == null);
```
hishamco: According to the function definition https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Framework.Localization.Abstractions/LocalizedString.cs#L30  I don't think that we need the named parameter ```resourceNotFound``` while you didn't change the order of the parameters

muratg:
I think we should do https://github.com/aspnet/Localization/issues/23 in this PR.

src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
DamianEdwards: We always tend to name args when we pass in `bool` or `null` values.

DamianEdwards:
@muratg I added changes for #23

DamianEdwards:
@hishamco that's what the extension method achieves https://github.com/aspnet/Localization/pull/22/files#diff-fd6180fc301d656a5a25bc6d20b88a50R38

hishamco:
:+1:

src/Microsoft.Framework.Localization.Abstractions/StringLocalizerExtensions.cs
```diff
@@ -10,22 +11,31 @@ public static class StringLocalizerExtensions
...
+        public static LocalizedString GetString(
+            [NotNull] this IStringLocalizer stringLocalizer,
+            [NotNull] string name,
+            params object[] arguments) => stringLocalizer[name, arguments];
+
+        /// <summary>
+        /// Gets all string resources including those for ancestor cultures.
+        /// </summary>
+        /// <param name="stringLocalizer">The <see cref="IStringLocalizer"/>.</param>
+        /// <returns></returns>
```
kirthik: `<returns>` should have some text. May be "The strings" like above?

kirthik:
:shipit:

muratg:
:shipit:

PR #24, opened 2015-05-21
Unnecessary using directives

hishamco:
Remove unnecessary directives from LocalizationSample and Localization source
/cc @DamianEdwards


hishamco:
@DamianEdwards Can you have a quick look to this PR?

PR #28, opened 2015-05-31
Simplify generating culture names

hishamco:
Simplify generating culture names using simple LINQ query with newly format style. 

For better performance we can fetch the culture names rather than fetching the entire ```CultureInfo``` list for all the cultures

### DIFF
--- a/src/CultureInfoGenerator/Program.cs
+++ b/src/CultureInfoGenerator/Program.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using Microsoft.Framework.Runtime;
 using Microsoft.Win32;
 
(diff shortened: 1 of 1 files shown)


src/CultureInfoGenerator/Program.cs
```diff
@@ -58,24 +59,7 @@ public static partial class CultureInfoCache
...
-                    {
-                        writer.WriteLine(",");
-                    }
-                    else
-                    {
-                        // Last entry
-                        writer.WriteLine();
-                    }
-                }
+                writer.WriteLine(string.Join(",\n", cultures.Select(c => $"            \"{c.Name}\"")));
```
Eilon: `\n` should probably be `Environment.NewLine`

hishamco: I do that to avoid using concatenation, but I can use string interpolation :)

DamianEdwards: I wouldn't worry about concatenation, this is a build-time tool we'll run once per Windows OS release. Optimize for code clarity, if anything at all.

Eilon: I think it's reasonable to argue that using `string.Join` is clearer than a `for()` loop.

DamianEdwards: Oh totally, not disputing that.

hishamco:
The PR has been update

Eilon:
@kirthik please review; the code looks good to me. It's just one of the build tools.

kirthik:
Looks good to me. I will merge it.

kirthik:
merged 59d957df3193902c9ee9beea0fe488b75fecee46

PR #32, opened 2015-06-16
return to the official xunit build

anurse:
Approved verbally for beta5 by @Eilon 

/cc @DamianEdwards @davidfowl @muratg


anurse:
We're going to take this to dev since it isn't actually needed to **build** release.

PR #37, opened 2015-06-18
Convert '_cookieSeparator' from char[] -> char

hishamco:


hishamco:
After discussion with @Eilon  in https://github.com/aspnet/Configuration/pull/225, ```_cookieSeparator``` should be ```char[]``` and this PR should be closed

PR #45, opened 2015-06-26
Make 'ResourceManagerStringLocalizer' overload internal

hishamco:
While we have ```InternalsVisibleTo``` in this line of https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Framework.Localization/Properties/AssemblyInfo.cs#L6, it will be nice to modify the access modifier for ```ResourceManagerStringLocalizer``` overload to ```internal``` rather than ```public``` because the intend for this overload to be used for testing purpose only

### DIFF
--- a/src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
+++ b/src/Microsoft.Framework.Localization/ResourceManagerStringLocalizer.cs
@@ -43,10 +43,7 @@ namespace Microsoft.Framework.Localization
             
         }
 
-        /// <summary>
-        /// Intended for testing purposes only.
-        /// </summary>
-        public ResourceManagerStringLocalizer(
+        internal ResourceManagerStringLocalizer(
             [NotNull] ResourceManager resourceManager,
             [NotNull] AssemblyWrapper resourceAssemblyWrapper,
             [NotNull] string baseName,


Eilon:
That ctor is for unit test purposes only - it's fine to take internal-ish parameters.

hishamco:
The issue is not with the internal-ish parameters, my doubt is why the constructor is public while the intention is used internally

Eilon:
So that it can be called from tests... It's documented as being for test-only scenarios.

hishamco:
Yes it's documented, but why we need to make it public while the intend is used for testing purpose, it doesn't make sense. IMHO ```internal``` is suited as access modifier

Eilon:
What is the harm in keeping it public?

hishamco:
It is not harmful, but meaning-wise it doesn't make sense to use ```public``` for testing only!!

Eilon:
I think it's fine to leave it the way it is. Otherwise we'd have hundreds of places in ASP.NET / EF / DNX where we'd need to go scan and change things, and I don't think that's worth it.

hishamco:
>Otherwise we'd have hundreds of places in ASP.NET / EF / DNX where we'd need to go scan and change things, and I don't think that's worth it.

Yep, but believe it or not it will consider as a lapse in the APIs and Framework Guidelines is obvious when shall we use ```public``` as well as ```internal```, I hope you understand what is my aim which is make ASP.NET 5 bits better than before, and nothing more :smile: 

All the best

PR #46, opened 2015-06-26
Remove extra line

hishamco:
I notice there are some files that have an extra line feed at the end, it will be nice if they are eliminated


Eilon:
Closing this because it has no real effect on the code. The problem with PRs that don't overall improve things is that they cause extra noise in notifications, make the CI queues longer, etc.

Issue #38, opened 2015-06-19
Fallback to wrong language preference when the top preference is unavailable


ChengTian:
I have the following configuration in my `Startup.cs`
```
var options = new RequestLocalizationOptions
{
    SupportedCultures = new List<CultureInfo>
    {
        new CultureInfo("zh-Hans"),
        new CultureInfo("en-US")
    }
};

app.UseRequestLocalization(options);
```
Then I set my language preference in IE with the following priorities (from highest to lowest):
1. ja (Japanese)
2. zh-Hans (Simplified Chinese)
3. en-US (English, United States)

However, the page fallback to en-US even when zh-Hans is available and has higher priority:

![capture](https://cloud.githubusercontent.com/assets/1383883/8263308/2fc9d2bc-168e-11e5-9761-6b7450181e97.PNG)

hishamco:
This is happen because if the requested culture is not available in the supported cultures that you already define then the current implementation is looking for the ```DefaultRequestCulture``` which is the ```CultureInfo.CurrentCulture```, BTW i'm trying to get a fix for this issue ASAP

hishamco:
@ChengTian can you try the fix that I made in my fork https://github.com/hishamco/Localization and let me know if it's solve the issue

hishamco:
@ChengTian I'm asking if the issue got fixed or not

ChengTian:
@hishamco , if you already test your changes, please send a PR. The owner of this repo will do some verification before merging your PR.

hishamco:
Ok @ChengTian, I will send it ASAP

hishamco:
I made this PR #50 to solve this issue

Eilon:
@kirthik can you re-verify the fix in https://github.com/aspnet/Localization/pull/50, merge it, and add any extra tests that we need?

DamianEdwards:
I actually think this should be fixed in `AcceptLanguageHeaderRequestCultureProvider`. Essentially, after it validates a culture, it should compare the result and if it's not using the culture that was passed in, it should move on to the next culture in the list (it already loops through the cultures, it just isn't doing the check).

https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNet.Localization/AcceptLanguageHeaderRequestCultureProvider.cs#L58

kirthik:
merged 7b53bdd43e2e605a8eb686f22ceb558f46a1c4f5

Eilon:
@kirthik mark as 3 - Done?

[...]

PR #50, opened 2015-07-07
Fix fallback language issue

hishamco:


src/Microsoft.AspNet.Localization/RequestCultureProvider.cs
```diff
@@ -41,12 +42,26 @@ protected RequestCulture ValidateRequestCulture(RequestCulture requestCulture)
 
             if (Options.SupportedCultures != null && !Options.SupportedCultures.Contains(result.Culture))
             {
-                result = new RequestCulture(Options.DefaultRequestCulture.Culture, result.UICulture);
+                if(Options.SupportedCultures.Count>0)
```
Eilon: Please fix the formatting in this file.

src/Microsoft.AspNet.Localization/RequestCultureProvider.cs
```diff
@@ -57,5 +72,17 @@ protected RequestCulture ValidateRequestCulture(RequestCulture requestCulture)
 
             return result;
         }
+        
+        private CultureInfo GetFallbackCulture()
```
Eilon: This is never called.

hishamco: You are right, I forget to remove it

src/Microsoft.AspNet.Localization/RequestCultureProvider.cs
hishamco: I didn't get you!! which formatting do you mean @Eilon

Eilon:
@DamianEdwards does this look good to you?

hishamco:
@DamianEdwards can you please verify if this PR fix the bug?

kirthik:
code change looks good to me but we should have tests to validate the scenario for `RequestCultureProvider` implementations.

hishamco:
That's what I wanna post ASAP, I prepare tests for ```RequestCultureProvider``` implementations

hishamco:
@kirthik can you validate from the scenario in issue #38 without applying my fix, seems it works fine now, i'm not sure if there's an issue in underlying Http APIs.
I wrote a unit test to validate that, I will try to push it today

DamianEdwards:
I'm not sure this is the right fix here. This change effectively means we now fallback to whatever culture is first in the supported list, rather than fallback to the configured default culture. If we allow configuration of a default I think we should fallback to that.

kirthik:
@hishamco closing this since @DamianEdwards said this fix is not right.

Issue #36, opened 2015-06-18
Make DefaultCookieName a static readonly field instead of a property


hishamco:
If we have look to this line of code https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNet.Localization/CookieRequestCultureProvider.cs#L47, It will be nice if declare ```DefaultCookieName``` as constant rather than read only property, because the value will never change.
```csharp

public static readonly string DefaultCookieName = "ASPNET_CULTURE";
```

hishamco:
\cc @dougbu

dougbu:
I'll leave this to triage to make a call.

hishamco:
\cc @DamianEdwards

hishamco:
@DamianEdwards can you have a look to this if you don't mind

hishamco:
Please let me know if a PR is needed for this change (if it's accepted)

kirthik:
cc @Eilon, @danroth27 for triaging.

Eilon:
This should remain a `static readonly` because it is not a universal truth, and it certainly *could* change.

hishamco:
Oops!! 
@Eilon that's why I did, have a look https://github.com/aspnet/Localization/pull/51/files#diff-fbf4e39686d45658872d8ef10043d143L47

Eilon:
Oh sorry, the bug and PR say `const` but the change is to make it `static readonly`.

hishamco:
No problem, but the tiltle is **Make DefaultCookieName a static readonly field instead of a property** perhaps the commit title :smile:

Eilon:
I just edited the title :smile:

hishamco:
lol BTW I meant be ```static readonly``` as constant :smile:

Eilon:
Ok, but just to be clear, `static readonly` is completely different from `const`.

hishamco:
Yep, but while the value will never change ```const``` will be equivalent with ```static readonly``` while the intend will never change the ```DefaultCookieName``` in the constructor

Eilon:
There is no guarantee that the cookie name won't change at some point in the future. We use `const` only for universal truths because the laws of the universe don't change (I hope!).

A `const` value gets baked into every usage, which means that if you later change the value of the `const` any already-compiled code will continue to use the old value. A `static readonly` is always established at first usage, and so the value can be changed in a newer build/version of the DLL.

Also, `const` values can be used in places where `const` values are required, such as in attribute values.

[...]

PR #51, opened 2015-07-07
Make DefaultCookieName a static readonly field instead of a property

hishamco:
I made this PR based on issue #36

### DIFF
--- a/src/Microsoft.AspNet.Localization/CookieRequestCultureProvider.cs
+++ b/src/Microsoft.AspNet.Localization/CookieRequestCultureProvider.cs
@@ -17,6 +17,11 @@ namespace Microsoft.AspNet.Localization
         private static readonly char[] _cookieSeparator = new[] { '|' };
         private static readonly string _culturePrefix = "c=";
         private static readonly string _uiCulturePrefix = "uic=";
+        
+        /// <summary>
+        /// Represent the default cookie name used to track the user's preferred culture information, which is "ASPNET_CULTURE".
+        /// </summary>
+        public static readonly string DefaultCookieName = "ASPNET_CULTURE";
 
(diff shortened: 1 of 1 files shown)


kirthik:
The corresponding issue for this PR is not triaged. I will wait for the triage to make a decision before looking at the PR.

Eilon:
I closed the issue - it's not something we should do.

hishamco:
Check the my last comment in the issue #36

src/Microsoft.AspNet.Localization/CookieRequestCultureProvider.cs
```diff
@@ -17,6 +17,10 @@ public class CookieRequestCultureProvider : RequestCultureProvider
         private static readonly char[] _cookieSeparator = new[] { '|' };
         private static readonly string _culturePrefix = "c=";
         private static readonly string _uiCulturePrefix = "uic=";
+        /// <summary>
```
Eilon: Add an empty line here.

hishamco: Do you mean after ```_uiCulturePrefix```? I already added what I suggest with the docs, i'm not sure if empty line is needed

Eilon: It looks weird to have a doc comment "touching" the field above it.

hishamco:
The PR has been updated
Thanks @Eilon

kirthik:
Looks good to me

Eilon:
@kirthik please merge.

kirthik:
merged c228b65d478561869daa9161fc97a89ff02cad2e

PR #56, opened 2015-07-22
Added repository information to project.json files

kichalla:
@kirthik

### DIFF
--- a/src/Microsoft.AspNet.Localization/project.json
+++ b/src/Microsoft.AspNet.Localization/project.json
@@ -1,24 +1,32 @@
 {
-  "version": "1.0.0-*",
-  "description": "Middleware for automatically applying culture information to HTTP requests.",
+    "version": "1.0.0-*",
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Localization/project.json
```diff
@@ -1,24 +1,32 @@
...
+            "dependencies": {
+                "Microsoft.CSharp": "4.0.0-beta-*",
+                "System.Collections": "4.0.10-beta-*",
+                "System.Globalization": "4.0.10-beta-*",
+                "System.Linq": "4.0.0-beta-*",
+                "System.Threading": "4.0.10-beta-*"
+            }
+        }
+    },
+
```
kirthik: remove extra line

kichalla: there are extra lines for other keys above, so kept it consistent...

PR #72, opened 2015-09-14
Fix for #38: Fallback to wrong language preference when the top preference is unavailable

kirthik:
cc @DamianEdwards, @Eilon


Eilon:
:shipit:

Issue #76, opened 2015-09-15
[Bug] Unable to run Localization Sample


hishamco:
When I run the localization sample I got the following exception **Could not resolve a service of type 'Microsoft.Framework.Localization.IStringLocalizer`1[[LocalizationSample.Startup, LocalizationSample, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]' for the parameter 'SR' of method 'Configure' on type 'LocalizationSample.Startup'.**

kirthik:
PR https://github.com/aspnet/Localization/pull/77 should fix that issue

hishamco:
Thanks @kirthik

kirthik:
merged 06aa241

PR #77, opened 2015-09-15
Update sample to use LocalizationOptions and fixed issue in AddLocalization

kirthik:
cc @Eilon


src/Microsoft.Framework.Localization/LocalizationOptions.cs
```diff
@@ -11,6 +11,6 @@ public class LocalizationOptions
         /// <summary>
         /// The relative path under application root where resource files are located.
         /// </summary>
-        public string ResourcesPath { get; set; }
+        public string ResourcesPath { get; set; } = string.Empty;
```
hishamco: While the default value is ```string.Empty```?!!

kirthik: By default we look for resources in the root folder of app. Hence it is empty.

hishamco: I see, thanks for clarification

hishamco:
:+1: 
I'm waiting for the fix, @kirthik I hope you close the issue #76 after you merge the PR

Eilon:
:shipit:

src/Microsoft.Framework.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -34,7 +34,7 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
             _resourcesRelativePath = localizationOptions.Value.ResourcesPath ?? string.Empty;
             if (!string.IsNullOrEmpty(_resourcesRelativePath))
             {
-                _resourcesRelativePath = _resourcesRelativePath.Replace("/", ".") + ".";
+                _resourcesRelativePath = _resourcesRelativePath.Replace("/", ".").Replace('\\', '.') + ".";
```
davidfowl: Why are we using the string overload and char overload on the same line? Also why not Path.DirectorySeparatorChar/Path.AltDirectorySeparatorChar? Is there a good reason to not use those?

hishamco: Can we use ```Regex.Replace(resourcesRelativePath, @"[\\|/]", ".")``` instead of create new object in each call of ```String.Replace```?

davidfowl: No do not use regex at all.

hishamco: I know it has a performance hit, but ```StringBuilder``` is a good choice too

davidfowl: Did you measure any of what you're suggesting?

hishamco: Not here, but according to what I did before ```StringBuilder``` is better than ```String``` specially in text manipulation.

davidfowl: Profiling might surprise you. We don't make performance changes without measurements to back them up.

kirthik: @davidfowl : I will change to `Path.DirectorySeparatorChar/Path.AltDirectorySeparatorChar`.

kirthik:
merged 06aa2412a879801bd85dfe369c6ad06a149b7cbb

PR #79, opened 2015-09-17
Enabling NuGetPackageVerifier

ajaybhargavb:
@Eilon

### DIFF
--- a/NuGetPackageVerifier.json
+++ b/NuGetPackageVerifier.json
@@ -1,0 +1,34 @@
+{
+    "adx": { // Packages written by the ADX team and that ship on NuGet.org
+        "rules": [
+            "AssemblyHasDocumentFileRule",
+            "AssemblyHasVersionAttributesRule",
+            "AssemblyHasServicingAttributeRule",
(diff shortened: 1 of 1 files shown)


NuGetPackageVerifier.json
```diff
@@ -0,0 +1,29 @@
...
+        "rules": [
+            "AssemblyHasDocumentFileRule",
+            "AssemblyHasVersionAttributesRule",
+            "AssemblyHasServicingAttributeRule",
+            "AssemblyHasNeutralResourcesLanguageAttributeRule",
+            "SatellitePackageRule",
+            "StrictSemanticVersionValidationRule"
+        ],
+        "packages": {
+            "CultureInfoGenerator": { },
```
Eilon: I think this one is non-shipping, no?

Eilon: And if it is, can remove the change to the assembly info for that package.

ajaybhargavb: yes, will update.

Eilon:
:shipit:

PR #81, opened 2015-09-17
Refomat html inside Startup.cs

hishamco:


Eilon:
Is this needed?

hishamco:
I know it's not much, but make thing pretty :smile:

Eilon:

hishamco:
Ok, thanks @Eilon

PR #82, opened 2015-09-17
Make 'reset' reset both Cookie & QueryString

hishamco:
Reset button should reset both cookie & query string instead of cookie only


Eilon:

PR #83, opened 2015-09-17
Sets 'en-US' as default culture in cultures list

hishamco:
The "-- select --" option enforce the culture to be invalid, it will be nice if we make the default culture  ```en-US```


Eilon:

Issue #88, opened 2015-09-22
[Bug] Omit the DefaultRequestCulture should not throw NullReferenceException


hishamco:
While I'm trying to fix the issue #70 I come up with the case that throws ```NullReferenceException``` when ```DefaultRequestCulture``` is not defined. The exception occurs from this line of code https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNet.Localization/AcceptLanguageHeaderRequestCultureProvider.cs#L60

Eilon:
@kirthik if the PR looks good to you, please resolve this issue.

kirthik:
How can `DefaultRequestCulture` be `null`? By default it is  set to current culture.

Eilon:
`ValidateRequestCulture` might return `null`?

hishamco:
In fact support both SupportedCultures & SupportedUICultures while omit the ```DefaultRequestCulture``` cause the error, I tested in my end. Please verify by running the test

kirthik:
merged 61106d36a17239d60ccd4cb6e0257a267d9f32ba

Issue #70, opened 2015-09-10
RequestCultureProvider sets an unsupported culture when the request is not supported and the default is not defined or not supported.


Bartmax:
Helping someone with an issue on Jabbr about localization, I found this strange (at least for me) behavior.

```c#
RequestLocalizationOptions localizationOptions = new RequestLocalizationOptions()
{
  SupportedCultures = new List<CultureInfo>
  {
    new CultureInfo("nl-NL")
  },
  SupportedUICultures = new List<CultureInfo>
  {
    new CultureInfo("nl-NL")
  }
  //, DefaultRequestCulture = new RequestCulture(new CultureInfo("nl-NL"), new CultureInfo("nl-NL"))
};
app.UseRequestLocalization(localizationOptions);

------------------------
services.AddLocalization();
services.AddMvc().AddViewLocalization();

```
**note the commented line**

With this cshtml (in layout or index):
```
<h1>UICulture: @System.Globalization.CultureInfo.CurrentUICulture.Name</h1> // outputs: en-US 
<h1>Culture: @System.Globalization.CultureInfo.CurrentCulture.Name</h1> // outputs: es-UY
```

According to the documentation here: https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNet.Localization/RequestLocalizationOptions.cs#L38-L54

This is not expected behavior. Those cultures are not supposed to be supported by the application.

Uncommenting the commented line (DefaultRequestCulture) I get the expected output:
```
DefaultRequestCulture = new RequestCulture(new CultureInfo("nl-NL"), new CultureInfo("nl-NL"))
UICulture: nl-NL
Culture: nl-NL
```

Looking at source code, I found that a 'premature' fallback to default culture.
https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNet.Localization/RequestCultureProvider.cs#L42-L50 without checking if the Default is supported.

I think the workflow should be more like (not real code)

```
var culture = Options.SupportedCultures == null || !Options.SupportedCultures.Any() || Options.SupportedCultures.Contains(result.Culture)
 	? result.Culture
	: Options.SupportedCultures.Contains(Options.DefaultRequestCulture.Culture)
	? Options.DefaultRequestCulture.Culture
	: Options.SupportedCultures.First();
	
var uiCulture =  Options.SupportedUICultures == null || !Options.SupportedUICultures.Any() || Options.SupportedUICultures.Contains(result.UICulture)
 	? result.UICulture
[...]

PR #89, opened 2015-09-22
Fix issue #70 and issue #88

hishamco:
I made this PR to fix the issues #70 and #88

### DIFF
--- a/src/Microsoft.AspNet.Localization/AcceptLanguageHeaderRequestCultureProvider.cs
+++ b/src/Microsoft.AspNet.Localization/AcceptLanguageHeaderRequestCultureProvider.cs
@@ -57,7 +57,7 @@ namespace Microsoft.AspNet.Localization
 
                         requestCulture = ValidateRequestCulture(requestCulture);
 
-                        if (requestCulture.Culture == culture)
+                        if (requestCulture != null && requestCulture.Culture == culture)
                         {
                             return Task.FromResult(requestCulture);
                         }
--- a/src/Microsoft.AspNet.Localization/AcceptLanguageHeaderRequestCultureProvider.cs
+++ b/src/Microsoft.AspNet.Localization/AcceptLanguageHeaderRequestCultureProvider.cs
@@ -57,7 +57,7 @@ namespace Microsoft.AspNet.Localization
 
                         requestCulture = ValidateRequestCulture(requestCulture);
 
-                        if (requestCulture != null && requestCulture.Culture == culture)
+                        if (requestCulture?.Culture == culture)
                         {
                             return Task.FromResult(requestCulture);
                         }
--- a/test/Microsoft.AspNet.Localization.Tests/AccessLanguageHeaderRequestCultureProviderTest.cs
+++ b/test/Microsoft.AspNet.Localization.Tests/AccessLanguageHeaderRequestCultureProviderTest.cs
@@ -106,5 +106,38 @@ namespace Microsoft.Framework.Localization.Tests
                 var response = await client.GetAsync(string.Empty);
             }
         }
+        
+        [Fact]
+        public async void OmitDefaultRequestCultureShouldNotThrowNullReferenceException_And_ShouldGetTheRightCulture()
+        {
+            using (var server = TestServer.Create(app =>
+            {
+                var options = new RequestLocalizationOptions()
+                {
+                    SupportedCultures = new List<CultureInfo>
+                    {
+                        new CultureInfo("ar-YE")
+                    },
+                    SupportedUICultures = new List<CultureInfo>
+                    {
+                        new CultureInfo("ar-YE")
+                    }
+                };
+                app.UseRequestLocalization(options);
+                app.Run(context =>
+                {
+                    var requestCultureFeature = context.Features.Get<IRequestCultureFeature>();
+                    var requestCulture = requestCultureFeature.RequestCulture;
+                    Assert.Equal("ar-YE", requestCulture.Culture.Name);
+                    Assert.Equal("ar-YE", requestCulture.UICulture.Name);
+                    return Task.FromResult(0);
+                });
+            }))
+            {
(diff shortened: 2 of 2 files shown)


Eilon:
@hishamco is this the full fix for #70? I thought the fix for that was quite different, no?

@kirthik can you please review? If this looks good, please resolve #88.

hishamco:
@Eilon regarding the issue #70 the original bug causes when the ```DefaultRequestCulture``` is not supported, this PR is fix this issue after the change made by @kirthik at the commit https://github.com/aspnet/Localization/commit/7b53bdd43e2e605a8eb686f22ceb558f46a1c4f5, i think we should create a new issue for another related scenarios that discussed to #70.
Please let me know if you have another suggestion @Eilon ?

Eilon:
Why not resolve issue #70 with a full fix for the scenario described there? What would the new issue say that isn't already said in #70?

The fix here seems fine for #88.

hishamco:
Yep it's fine for issue #88, but if you have a look to the first post in the issue #70 uncomment the following line
```csharp
//, DefaultRequestCulture = new RequestCulture(new CultureInfo("nl-NL"), new CultureInfo("nl-NL"))
```
getting the excepting result, which is done by this PR after the change in the commit https://github.com/aspnet/Localization/commit/7b53bdd43e2e605a8eb686f22ceb558f46a1c4f5, other posts raised other issues related to the original bug. i hope you understand what I mean

src/Microsoft.AspNet.Localization/AcceptLanguageHeaderRequestCultureProvider.cs
```diff
@@ -57,7 +57,7 @@ public override Task<RequestCulture> DetermineRequestCulture([NotNull] HttpConte
 
                         requestCulture = ValidateRequestCulture(requestCulture);
 
-                        if (requestCulture.Culture == culture)
+                        if (requestCulture != null && requestCulture.Culture == culture)
```
davidfowl: `requestCulture?.Culture == culture`

hishamco: Looks good :smile: , thanks @davidfowl

kirthik:
I don't understand how this change is touching #70. Can you please edit PR title so it says it just fixes #88?

test/Microsoft.AspNet.Localization.Tests/AccessLanguageHeaderRequestCultureProviderTest.cs
```diff
@@ -106,5 +106,38 @@ public class AccessLanguageHeaderRequestCultureProviderTest
                 var response = await client.GetAsync(string.Empty);
             }
         }
+        
+        [Fact]
+        public async void OmitDefaultRequestCultureShouldNotThrowNullReferenceException_And_ShouldGetTheRightCulture()
```
kirthik: How is this test different from `GetFallbackLanguage_ReturnsFirstNonNullCultureFromSupportedCultureList`. Except for the added `SupportedUICultures` list in this test, I don't see any difference.

hishamco:
@kirthik the main difference between ```OmitDefaultRequestCultureShouldNotThrowNullReferenceException_And_ShouldGetTheRightCulture``` and ```GetFallbackLanguage_ReturnsFirstNonNullCultureFromSupportedCultureList``` is https://github.com/aspnet/Localization/blob/dev/test/Microsoft.AspNet.Localization.Tests/AccessLanguageHeaderRequestCultureProviderTest.cs#L55

test/Microsoft.AspNet.Localization.Tests/AccessLanguageHeaderRequestCultureProviderTest.cs
hishamco: @kirthik the main difference between OmitDefaultRequestCultureShouldNotThrowNullReferenceException_And_ShouldGetTheRightCulture and GetFallbackLanguage_ReturnsFirstNonNullCultureFromSupportedCultureList is https://github.com/aspnet/Localization/blob/dev/test/Microsoft.AspNet.Localization.Tests/AccessLanguageHeaderRequestCultureProviderTest.cs#L55

kirthik: The line you pointed to is `GetFallbackLanguage_ReturnsFromSupportedCulture_AcceptLanguageListContainsSupportedCultures`. I am talking about `GetFallbackLanguage_ReturnsFirstNonNullCultureFromSupportedCultureList`

hishamco:
@kirthik the main issue in #70 that when we omit the ```DefaultRequestCulture``` the ```CurrentCulture``` & ```CurrentUICulture``` will be **es-US**, this PR fix it, i know there are some scenarios related to issue #70, if you see the rename of the title is a proper choice i will do it now

test/Microsoft.AspNet.Localization.Tests/AccessLanguageHeaderRequestCultureProviderTest.cs
hishamco: Oops .. sorry, seems introducing ```SupportedUICultures``` property cause the error
```csharp
SupportedUICultures = new List<CultureInfo>
{
    new CultureInfo("ar-YE")
}
```
Can you check please ..

hishamco: In fact support both ```SupportedCultures``` & ```SupportedUICultures``` cause the error, I tested in my end. If you notice not ```OmitDefaultRequestCultureShouldNotThrowNullReferenceException_And_ShouldGetTheRightCulture``` is the only test that support both of them

kirthik:
merged 61106d36a17239d60ccd4cb6e0257a267d9f32ba

PR #101, opened 2015-10-07
Update culture names list

hishamco:
I create this PR based on issue #98

PR #103, opened 2015-10-08
Fix to google chrome with the header zh-Hans-CN

Kagamine:
The most of browsers will detect the accept lang into zh or zh-CN, but in chrome it will be zh-Hans-CN.
By the way, I am not suggest put cultures in a class, why not put them into the resx? I think a resource file contains the cultures which supported in the file is better.


hishamco:
@Kagamine the list already updated in the PR #101 , furthermore the list is updated by running ```CultureInfoGenerator``` command not by adding entries directly to the list

thanks

hishamco:
Your thoughts @DamianEdwards

Kagamine:
There were some wrong with PR #101 with line 823 824

```
+            "zh-Hans-HK",
+            "zh-Hans-MO",
```

It should be zh-Hant-HK and zh-Hant-MQ and missing zh-Hans-CN zh-Hant-TW

Kagamine:
I will close this pr and make a new. there were something wrong with it.

hishamco:
@kagamine  these culture generated by tool not by me :smile:

PR #107, opened 2015-10-09
Add all localized strings for zh

Kagamine:
@hishamco Done.

### DIFF
--- a/samples/LocalizationSample/My/Resources/Startup.zh-CN.resx
+++ b/samples/LocalizationSample/My/Resources/Startup.zh-CN.resx
@@ -117,7 +117,79 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="Big number (invariant):" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


kirthik:
@Kagamine : As I mentioned in the other PR, either we should leave the code as is or for completeness update all resx files. Updating just one locale will raise more questions from other users about why one locale was localized while others are not. 

If you are really interested in pushing your PR, please use a translation tool to localize strings in all resx files.

hishamco:
+1 @kirthik , nothing but while the intend in to show the localized value for the key, ```Hello``` key is enough IMHO

Kagamine:
I think the translate tool will make some mistake with these string, hishamco did that in another PR, and his translated zh-cn strings are wrong.

hishamco:
@Kagamine the mistake in Chinese translation was in the commit https://github.com/aspnet/Localization/commit/a5a264d1ad12e55dfbcb569563165ab4475a1d01 which is not created by me :smile: anyhow you can fix it to the correct value, but there's not much value. there are many PRs closed by @Eilon for the same reason, because it's a sample nothing else

Eilon:
I think the `Hello` key is enough because that's the only example needed to show string localization. The purpose of this sample is to show one example of each type of loc/glob concept, which is already does.

If any of the existing translations of the "Hello" string are incorrect, then those should be fixed.

Issue #111, opened 2015-10-14
Design changes to known/allowed culture list and cache


Eilon:
1. Have default allowed culture list be empty
 * Make setting the default culture mandatory when using AddLocalization
 * If you want to support a specific set of cultures, pass in a default culture and also a set of allowed cultures
2. The loc middleware upon seeing the request’s desired cultures will compare it against the list of SupportedXyz cultures for an exact match (case insensitive). If it’s not any of those, fall back to the default culture
3. Rename AllowedXyz to SupportedXyz for lists of cultures
4. Keep CultureInfoCache and use it for all CultureInfo retrieval, including registering “unknown” cultures. The list of Known Cultures will go away.

omuleanu:
@Eilon we should have the option to make the application use the browser language (Accept-Language header) without having to specify every single possible culture in `RequestLocalizationOptions.SupportedCultures`

Eilon:
@omuleanu I believe there were some concerns around allowing completely arbitrary cultures that browsers send, which would potentially cause internal culture caches to potentially grow infinitely, causing memory issues. On OSes such as Windows 10, there is an infinite number of valid cultures, and ASP.NET Core Localization has to cache them (because creating cultures is a bit expensive), so if browsers send en-US-1, en-US-2, ..., en-US-9999 then ASP.NET's culture cache would grow and take up all memory.

Our belief is that sites only support a relatively small set of cultures (at most a few dozen, and that's extremely rare), so listing them all is reasonable.

If a site wishes to support completely arbitrary cultures (i.e. allow *anything* a browser sends), then they will have to use some other middleware.

@ryanbrandenburg @DamianEdwards was this the exact reason? Or was there something else?

PR #113, opened 2015-10-16
Fix issue #70

hishamco:
I made this PR to fix issue #70 , I added the a unit tests to cover the following cases:
* If the request is not supported and ```DefaultRequestCulture``` is not defined
* If the request is not supported and ```DefaultRequestCulture``` is not supported
* If the request is not supported and ```SupportedCultures``` & ```SupportedUICultures``` are not defined
* If the request is not supported and ```SupportedCultures``` & ```SupportedUICultures``` are empty

/cc @kirthik @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Localization/RequestCultureProvider.cs
+++ b/src/Microsoft.AspNet.Localization/RequestCultureProvider.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved. 
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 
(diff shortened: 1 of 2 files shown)


hishamco:
@kirthik can you please review this?

kirthik:
With this design change - https://github.com/aspnet/Localization/issues/111, passing in default culture will be required when adding localization. We will scan request's culture through supported culture list and fall back to default if exact match is not found. So scenario where default culture is not defined or supported does not arise. I am not sure if this PR is relevant anymore.

hishamco:
I don't know why we should enforce the default culture while it's **en-US** in previous releases of ASP.NET?
If the PR irrelevant is that mean we should close issue #70 now?!!

Eilon:
#70 is probably not entirely needed because one of the reasons for the issue is that the default culture is optional. Once it's non-optional, most (or all?) of the issue goes away.

kirthik:
@Eilon: that's right. We can close this PR and https://github.com/aspnet/Localization/issues/70

Eilon:
@kirthik do as you wish :smile:

PR #114, opened 2015-10-16
fix #90: ResourceStringLocalizerFactory creates string localizer based on resourcetype's name instead of full name

### LINKED ISSUE #90: ResourceStringLocalizerFactory creates string localizer based on resourcetype's name instead of full name
https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs#L53

We should use resourceName.FullName so that if same class exists under different namespaces, we can have different `.resx` for each type.

kirthik:
cc @DamianEdwards, @Eilon


src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -64,7 +64,16 @@ public IStringLocalizer Create(Type resourceSource)
 
             var typeInfo = resourceSource.GetTypeInfo();
             var assembly = typeInfo.Assembly;
-            var baseName = _applicationEnvironment.ApplicationName + "." + _resourcesRelativePath + resourceSource.Name;
+
+            string baseName = string.Empty;
+            if (!string.IsNullOrEmpty(_resourcesRelativePath))
```
hishamco: Can you use this
```csharp
baseName = $"{_applicationEnvironment}.{_resourcesRelativePath}{typeInfo.FullName}";
```
instead, because if the ```resourcesRelativePath``` is null or empty the value will be nothing. I used a string interpolation to make things pretty :smile:

hishamco:
Nice work @kirthik

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
Eilon: I would do neither. String interpolation adds no value here, and the `if()` is indeed not necessary. I'd just do `baseName = _applicationEnvironment.ApplicationName + "." + _resourcesRelativePath + typeInfo.FullName;` always.

Eilon: Which, looking at the code, is I guess already what it was...

Eilon:
After the small comment, :shipit:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
hishamco: I didn't see that :smile:

kirthik: ya..not sure why I replaced the code :) thanks for catching it.

kirthik:
merged 90116ddd78a8a9f3e5e5ba68500b799e9dc3ec8b

PR #129, opened 2015-10-24
Change dnx451 -> net451

hishamco:
React to the change in commit https://github.com/aspnet/Localization/commit/c6bf98c9b7be4641282fe26a13972fbd02100c4b
BTW why the changes doesn't include the samples and unit tests?
/cc @pranavkm @kirthik


hishamco:
This no longer needed after PR #130

PR #137, opened 2015-10-31
Add missing wwwroot folder

hishamco:
From awhile I'm working with Visual Studio Code and Command Line, today I used a Visual Studio to debug the localization sample, unfortunately an error message show up, that **wwwroot** is not found. So adding the missing folder will fix the issue
/cc @kirthik


hishamco:
I will close this after @DamianEdwards commit e93a00c2c4d3574f169215eea214901af35a96ee

hishamco:
@DamianEdwards does we need **Web.Config** again?!! while the new configurations APIs introduced

PR #145, opened 2015-11-10
Run CoreCLR test when runtime framework is CoreCLR

troydai:
This change fixes test break on Mac. In short, the test requires deploying a server on Mono shouldn't be run when the current runtime framework is CoreCLR. Because in that case the path to the mono runtime can't be found on `$PATH`.

The change simply split one `theroy` into two and run under different runtime. The build script is responsible of running test on each framework, which it does today.

/cc @kirthik @cesarbs


cesarbs:
:shipit: when it passes on Travis.

troydai:
Travis fails because other reasons. I'll merge this first and get back to address the travis failure.

Issue #105, opened 2015-10-08
Localized string storage, Culture set, Front-end supports.


Kagamine:
/cc @hishamco @DamianEdwards @kirthik 

- To make a ILocalizedStringProvider and allow different ways to storage the string, such as .json like i18n did, or store these in EF. It should be writable by using the provider instance. Resx is so heavy and has a bad readability, it is hard to edit by using vs code or other editors.
- Make a culture set, to describe a set of localized strings, and these string will be displayed when the set contains the target culture. For this, we do not need CultureList.cs anymore.
- Make a reload or refresh function in the above interface, and we can use Reload() to reload strings when someone made some changes external. Or implement a resource listener.
- Add a middle ware to map a javascript to ensure we can use these string in front-end.
- Implement a RouteRequestCultureProvider if possible.

I did some works to implement these, but not based on this repo, I am too fool and cannot understand the structure of this repo. If you think my suggestions are valuable, you can have a look https://github.com/CodeComb/Localization

EF-Storage: https://github.com/CodeComb/Localization/blob/dev/src/CodeComb.AspNet.Localization/EntityFramework/EFCollection.cs

Json-Storage: https://github.com/CodeComb/Localization/blob/dev/src/CodeComb.AspNet.Localization/Json/JsonCollection.cs

Json sample:
https://github.com/CodeComb/Localization/blob/dev/test/CodeComb.AspNet.Localization.Tests/Localization/en-us.json

Javascript Localizer: https://github.com/CodeComb/Localization/blob/dev/src/CodeComb.AspNet.Localization/Extensions/LocalizationScriptMiddleware.cs

hishamco:
@Kagamine If i'm not wrong I saw an issue related to what you said before. The guy using the localization repo to store the key/value in database instead of resx file.
I will look for that issue and get back to you

danroth27:
We should look at providing some more samples that shows how to use alternative localized resource providers, like a DB.

hishamco:
@danroth27 should this will be included in the docs or just in the sample?
If there's a possibility to send a PR for this I will do ..

danroth27:
@Eilon is going to look at putting a sample together first and then we can look at documenting the recommended patterns.

Kagamine:
What's your view about the other lines in this issue such as frontend support? @danroth27

hishamco:
I was thinking to send a PR for this, but if @Eilon will create the sample, this is fine

hishamco:
[...]

PR #149, opened 2015-11-16
Add EntityFramework Localization Sample

hishamco:
I see many users asking about localization sources, and how to use other sources such as json, database .. etc instead of resx files. I provide this example to use EntityFramework as source


hishamco:
Can we consider this a helpful sample, that will give the developers a glimpse to how extend localization APIs to create a various data sources 
/cc @kirthik @Eilon

davidfowl:
This sorta thing belongs in Entropy.

hishamco:
@davidfowl but there are many question that i saw about change localization store, this issue #105 that let me take the initiative to provide such a sample in this repo

davidfowl:
@hishamco That's great but I'm asking you to move it to Entropy. Where all of our other examples like this belong.

hishamco:
ok @davidfowl I will move there, and i will mention it here
thanks

hishamco:
Based on @davidfowl suggestion, i moved the PR to https://github.com/aspnet/Entropy/pull/76

Issue #156, opened 2015-12-02
Change ConcurrentDictionary in ResourceManagerStringLocalizer to a regular dictionary


Eilon:
See: https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs#L21-L22

This code uses a `ConcurrentDictionary` even though the code is run on a single thread. It was once part of some static caching system, but that is long since gone, so it can safely be changed to a regular dictionary.

This was found while looking at https://github.com/aspnet/Localization/issues/142.

DamianEdwards:
So it turns out that this member **really** needs to be on the factory, as the performance once a page is localized right now is absolutely terrible (like, >6s to load the page). Making this cache global returns it to sane numbers after the initial render (<20ms).

ryanbrandenburg:
So @DamianEdwards you're saying that part of making that cache a singleton will be that this has to stay a ConcurrentDictionary and therefore this bug is obsolete?

Eilon:
@ryanbrandenburg I'm guessing the cache is just in the wrong place, maybe? As in, instead of being on the localizer, which is transient (so caching is almost useless), the cache needs to move to the factory, which is a singleton.

DamianEdwards:
Yeah I have a prototype, I'll send a PR.

DamianEdwards:
Fixed in 96462b9528da1b01d5f1759f28e49f2fb7cb2d7d

PR #160, opened 2015-12-08
Make the ResourceManagerStringLocalizerFactory cache localizer instances

DamianEdwards:
Turns out the performance of the `ResourceManagerStringLocalizer` is terrible in an actual app due to the way `ResourceManager` works. You really need to keep instances of `ResourceManager` around forever as it does a lot of its own caching internally and the existing localization experience in .NET apps today is designed around this premise.

The simplest change was to just make the `ResourceManagerSringLocazlierFactory` cache the instances of `ResourceManagerStringLocalizer` it creates, as they hold the `ResourceManager` inside of themselves, as well as a cache of missing resource manifest names.

This change gets us to a place where the `ResourceManager` instances are used more like they were intended to be, as well as the cache of missing resource lookups being effectively global. Performance gains are huge, e.g. localized home page from starter web template goes from >6s per page load to <20ms (that's not a typo, 6,700ms to 15ms, yes the `ResourceManager` **really** needs to be cached).

- #156

@Eilon @ryanbrandenburg

### DIFF
--- a/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
+++ b/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
@@ -2,11 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
 using System;
+using System.Collections.Concurrent;
 using System.IO;
 using System.Reflection;
 using System.Resources;
-using Microsoft.Extensions.PlatformAbstractions;
 using Microsoft.Extensions.OptionsModel;
+using Microsoft.Extensions.PlatformAbstractions;
 
 namespace Microsoft.Extensions.Localization
 {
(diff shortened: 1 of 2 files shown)


hishamco:
You can remove unnecessary ```using System.Collections.Concurrent;``` in **ResourceManagerStringLocalizer.cs** and :+1: for this PR

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -7,6 +7,7 @@
 using System.Resources;
 using Microsoft.Extensions.PlatformAbstractions;
 using Microsoft.Extensions.OptionsModel;
+using System.Collections.Concurrent;
```
Eilon: And sort 'em, too! :smile:

Eilon:
:shipit:

DamianEdwards:
96462b9528da1b01d5f1759f28e49f2fb7cb2d7d

PR #168, opened 2015-12-11
Make ResourceManagerStringLocalizer thread safe

DamianEdwards:
Make ResourceManagerStringLocalizer thread safe now instances are cached & shared

@Eilon @ryanbrandenburg


rynowak:
:shipit:

Issue #164, opened 2015-12-10
Improve usability of the request localization middleware


DamianEdwards:
- [x] Remove the `defaultRequestCulture` parameter from `IApplicationBuilder.UseRequestLocalization` and make it a property on `RequestLocalizationOptions` instead
- [x] Have just a single overload for `IApplicationBuilder.UseRequestLocalization` that takes the options

Issue #127, opened 2015-10-23
Add a builder API for configuring UseRequestLocalization


rynowak:
This would be more inline with what we do in other places in the framework. The use the current experience you need to add 2 usings that you may not likely have in your startup code.

**Suggested**

```
app.UseRequestLocalization(options => options.AddCultures("en-US", "en-GB", "fr").SetDefault("en-US"));
```

**Current**

```
using System.Collections.Generic;
using System.Globalization;
using Microsoft.AspNet.Localization;

            var localizationOptions = new RequestLocalizationOptions()
            {
                SupportedCultures = new List<CultureInfo>()
                {
                    new CultureInfo("en-US"),
                    new CultureInfo("en-GB"),
                    new CultureInfo("fr"),
                },
                SupportedUICultures = new List<CultureInfo>()
                {
                    new CultureInfo("en-US"),
                    new CultureInfo("en-GB"),
                    new CultureInfo("fr"),
                },
            };

            app.UseRequestLocalization(options, new RequestCulture(new CultureInfo("en-US")));
```

hishamco:
Great, that is what i wanna suggest but with little modification
```csharp
app.UseRequestLocalization(options => {
    options.SupportedCultures("en-US", "en-GB", "fr")
           .SupportedUICultures("en-US")
           .Default("en-US")
});
```
IMHO Fluent API in such case is a proper option

Eilon:
The builder APIs need to have verbs in them, so `AddSupportedCultures` works, but not just `SupportedCultures`.

hishamco:
So something like this will be fine
```csharp
app.UseRequestLocalization(options => {
    options.AddSupportedCultures("en-US", "en-GB", "fr")
           .AddSupportedUICultures("en-US")
           .SetDefault("en-US")
});
```

Eilon:
@kirthik @rynowak I'd like to spend a few minutes designing this next week so we agree on the pattern.

DamianEdwards:
Just going to change this to use the options setter delegate pattern instead, e.g.:
``` C#
app.UseRequestLocalization(options =>
	options.DefaultRequestCulture = new RequestCulture("en-US");
	options.SupportedCultures = new [] { "en-US", "fr" };
	options.SuppoertedUICultures = options.SupportedCultures;
);
```

hishamco:
[...]

PR #170, opened 2015-12-11
Improve usability of the request localization middleware #164

DamianEdwards:
#164 

Fixed up some out of date doc comments too.

@rynowak @ryanbrandenburg @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Localization/ApplicationBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Localization/ApplicationBuilderExtensions.cs
@@ -14,44 +14,14 @@ namespace Microsoft.AspNet.Builder
     {
         /// <summary>
         /// Adds the <see cref="RequestLocalizationMiddleware"/> to automatically set culture information for
-        /// requests based on information provided by the client using the default options.
-        /// </summary>
-        /// <param name="app">The <see cref="IApplicationBuilder"/>.</param>
-        /// <param name="defaultRequestCulture">The default <see cref="RequestCulture"/> to use if none of the
-        /// requested cultures match supported cultures.</param>
-        /// <returns>The <see cref="IApplicationBuilder"/>.</returns>
-        public static IApplicationBuilder UseRequestLocalization(
-            this IApplicationBuilder app,
-            RequestCulture defaultRequestCulture)
-        {
-            if (app == null)
-            {
-                throw new ArgumentNullException(nameof(app));
-            }
-
-            if (defaultRequestCulture == null)
-            {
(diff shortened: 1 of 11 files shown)


hishamco:
I'm not sure if the issue #127 should take place here or will be ignored

src/Microsoft.AspNet.Localization/RequestLocalizationOptions.cs
```diff
@@ -25,20 +29,40 @@ public RequestLocalizationOptions()
...
+        {
+            get
+            {
+                return _defaultRequestCulture;
+            }
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException();
```
ryanbrandenburg: Should have the name of the property in the exception yeah?

src/Microsoft.AspNet.Localization/ApplicationBuilderExtensions.cs
```diff
@@ -63,12 +33,7 @@ public static class ApplicationBuilderExtensions
                 throw new ArgumentNullException(nameof(options));
             }
 
-            if (defaultRequestCulture == null)
```
ryanbrandenburg: We checked that we didn't set the DefaultRequestCulture to null, but are we checking that it has ever been set anywhere or did I miss it?

DamianEdwards: It's initialized directly via the backing field https://github.com/aspnet/Localization/pull/170/files#diff-61f74221a354a19742c79b0c32c46423R15

src/Microsoft.AspNet.Localization/RequestLocalizationOptions.cs
DamianEdwards: Yeah I actually wasn't sure what name to use here, but property name makes sense.

src/Microsoft.AspNet.Localization/ApplicationBuilderExtensions.cs
ryanbrandenburg: So it is!

DamianEdwards:
@hishamco that will be separate

src/Microsoft.AspNet.Localization/RequestLocalizationOptions.cs
rynowak: This should be `nameof(value)`. That is the name of the parameter.

ryanbrandenburg:
:shipit:

Issue #171, opened 2015-12-13
[Bug] Unable to run Microsoft.Extensions.Localization.Tests


hishamco:

DamianEdwards:
![image](https://cloud.githubusercontent.com/assets/249088/11802367/bf9bc0ba-a2a4-11e5-894c-119c9606f946.png)

And on the CI: https://travis-ci.org/aspnet/Localization

You'll need to provide more detail.

hishamco:
What it made me crazy that it runs well on the CI, have a look to the exception
![capture](https://cloud.githubusercontent.com/assets/3237266/11810445/3ce336da-a33f-11e5-8171-0d40cf2e6b95.JPG)
```IOptionsMonitor<>``` not found, because there's no reference in the project.json

DamianEdwards:
can you try `git clean -xdf` in the repo root then run `build.cmd`

hishamco:
It works :smile: 
thanks

PR #172, opened 2015-12-13
Add missing 'Microsoft.Extensions.OptionsModel' namepace

hishamco:
After the latest commits, there's a namespace lost by mistake, so this fixes issue #171 
/cc @DamianEdwards @Eilon


DamianEdwards:
Why do you say it's missing? If that was missing from the project.json file, it wouldn't compile.

hishamco:
Yep, it's not compile at my end, after I pull the latest changes

Issue #157, opened 2015-12-02
Append ApplicationName and RelativePath doesn't make sense for types from class library in ResourceManagerStringLocalizerFactory


martinba1:
See:
https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs#L72

It is useful to append _applicationEnvironment.ApplicationName, _resourcesRelativePath to baseName only for types from the main application assembly. It makes impossible to use this factory for resources shared in class library assemblies.

Eilon:
If the type name being requested doesn't start with the app name then we shouldn't use any fancy prefix logic. This way the application's configuration for localization won't interfere with class libraries doing their own localization.

joeaudette:
this is kind of a bad bug, only solution at the moment is to fork ResourceManagerStringLocalizerFactory and change the logic, wondering why this is backlog, thought it would be fixed in rc2 but backlog suggests it won't even be fixed in RTM

For example if I have a class library with a FooController like this:

    using Microsoft.AspNetCore.Mvc;
    using Microsoft.Extensions.Localization;
	namespace WebLib.Controllers
	{
		public class FooController : Controller
		{
			public FooController(
				IStringLocalizerFactory localizerFactory
				)
			{
				_localizer = localizerFactory.Create("MyResources.Controllers.FooController", "WebLib"); ;
			}

			private readonly IStringLocalizer _localizer;

			public IActionResult Index()
			{
				ViewData["Title"] = _localizer["Foo"];
				ViewData["Message"] = _localizer["Greetings ya'll"];

				return View();
			}
		}
	}

	and in the project.json I embed the MyResources folder which contains Controllers.FooController.fr.resx
	
	"buildOptions": {
        "embed": [ "Views/**", "MyResources/**" ]
    },
	
It works if the main web app is configured like this:

    services.AddLocalization();
	
but as soon as the web app decides to use this configuration:

    services.AddLocalization(options => options.ResourcesPath = "Resources" );
	
It breaks the localization in FooController even though I'm using the factory and asking for exactly the correct name of my resource

joeaudette:
and a further note, it is currently not possible in a class library to get a working IStringLocalizer like this:

    using Microsoft.AspNetCore.Mvc;
    using Microsoft.Extensions.Localization;

    namespace WebLib.Controllers
    {
        public class OtherController : Controller
        {
            public OtherController(
                IStringLocalizer<OtherController> localizer
[...]

PR #173, opened 2015-12-14
ResourcePath Empty

### LINKED ISSUE #138: Resource name is incorrect when not setting Options.ResourcesPath
See the line at https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs#L68

If `RequestLocalizationOptions.ResourcesPath` is **not** set (the default) then this will effectively cause the application name to be appended to the beginning of the full type name, thus making it impossible to actually resolve that resource.

This means you can't just create a resx file next to a class' cs file with the same name and have it work, e.g. `HomeController.cs` and `HomeController.en-AU.resx`

The line should probably be something like:
``` C#
var baseName = string.IsNullOrEmpty(_resourcesRelativePath)
                ? typeInfo.FullName
                : _applicationEnvironment.ApplicationName + "." + _resourcesRelativePath + typeInfo.FullName;
```

ryanbrandenburg:
If the ResourcePath is left empty it should still do something sane (default to checking for the resource next to the corresponding cs file).

Fixes #138


hishamco:
@ryanbrandenburg is issue #157 should go along with this PR or not

ryanbrandenburg:
@hishamco though the two are related by their location in the code they're looking to accomplish very different things and I think should be discussed and worked on each by themselves.

ryanbrandenburg:
@DamianEdwards for review.

hishamco:
:+1: after the unit test fix

DamianEdwards:
:shipit:

Issue #112, opened 2015-10-14
Fall back to parent cultures from the supported list when the desired culture is not available or known


Eilon:
Have loc selectors fall back to parent cultures from the supported list when the desired culture is not available or known. E.g. if the request culture is “en-Damian” and the supported cultures has “en”, then fall back to that.

hishamco:
What if the supported cultures have "en-US" should fallback to that or return ```null```?

Eilon:
@hishamco that depends on what the supported cultures are *and* on what the *requested* culture is. Can you give a more complete example?

BTW, either way we don't plan on doing this work item now - it's on the Backlog.

miloush:
@hishamco `null` would be the current behavior in .NET resources fallback and there might be sensitive cases where you don't want to mix it up. I agree that fallback to the parent culture is an important feature, though, and probably provider specific (if you set the cookie, set something you don't need fallback for).

DamianEdwards:
Going to make this depth-first instead so re-opening.

PR #174, opened 2015-12-15
Added support for hierarchical culture fallback

DamianEdwards:
- Enabled by default and configred by `RequestLocalizationOptions.FallbackToAncestorCulture`/`FallbackToAncestorUICulture`
- Tries all candidate cultures first before trimming the list to parents and trying again (breadth first), until a match is found or none is found
- Updated functional tests to cover fallback case
- #112

@ryanbrandenburg @Eilon @rynowak @lodejard

### DIFF
--- a/src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
+++ b/src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
@@ -18,6 +18,8 @@ namespace Microsoft.AspNet.Localization
     /// </summary>
     public class RequestLocalizationMiddleware
     {
+        private static readonly int MaxCultureFallbackDepth = 5;
+
         private readonly RequestDelegate _next;
         private readonly RequestLocalizationOptions _options;
 
@@ -27,9 +29,7 @@ namespace Microsoft.AspNet.Localization
         /// <param name="next">The <see cref="RequestDelegate"/> representing the next middleware in the pipeline.</param>
         /// <param name="options">The <see cref="RequestLocalizationOptions"/> representing the options for the
         /// <see cref="RequestLocalizationMiddleware"/>.</param>
-        public RequestLocalizationMiddleware(
-            RequestDelegate next,
-            RequestLocalizationOptions options)
+        public RequestLocalizationMiddleware(RequestDelegate next, RequestLocalizationOptions options)
         {
             if (next == null)
             {
@@ -75,22 +75,32 @@ namespace Microsoft.AspNet.Localization
                         CultureInfo uiCultureInfo = null;
                         if (_options.SupportedCultures != null)
                         {
-                            cultureInfo = GetCultureInfo(cultures, _options.SupportedCultures);
+                            cultureInfo = GetCultureInfo(
+                                cultures,
+                                _options.SupportedCultures,
+                                _options.FallbackToAncestorCulture,
+                                currentDepth: 0);
                         }
 
                         if (_options.SupportedUICultures != null)
                         {
-                            uiCultureInfo = GetCultureInfo(uiCultures, _options.SupportedUICultures);
+                            uiCultureInfo = GetCultureInfo(
+                                uiCultures,
+                                _options.SupportedUICultures,
+                                _options.FallbackToAncestorUICulture,
+                                currentDepth: 0);
                         }
 
                         if (cultureInfo == null && uiCultureInfo == null)
                         {
                             continue;
                         }
+
                         if (cultureInfo == null && uiCultureInfo != null)
                         {
                             cultureInfo = _options.DefaultRequestCulture.Culture;
                         }
+
                         if (cultureInfo != null && uiCultureInfo == null)
                         {
                             uiCultureInfo = _options.DefaultRequestCulture.UICulture;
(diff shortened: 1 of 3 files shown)


hishamco:
FYI don't miss to react to @ryanbrandenburg unit test changes in #173 unless you will commit first

src/Microsoft.AspNet.Localization/RequestLocalizationOptions.cs
```diff
@@ -51,6 +51,33 @@ public RequestCulture DefaultRequestCulture
...
+        /// culture determined by the configured <see cref="IRequestCultureProvider"/>s is not in the
+        /// <see cref="SupportedCultures"/> list but an ancestor culture is.
+        /// Defaults to <c>true</c>;
+        /// </summary>
+        /// <example>
+        /// If this property is <c>true</c> and the application is configured to support the culture "fr", but not the
+        /// culture "fr-FR", and a configured <see cref="IRequestCultureProvider"/> determines a request's culture is
+        /// "fr-FR", then the request's culture will be set to the culture "fr", as it is an ancestor of "fr-FR".
+        /// </example>
+        public bool FallbackToAncestorCulture { get; set; } = true;
```
ryanbrandenburg: Is ancestor the word they use in the standard or something? I would have thought of it as a "Parent Culture". Not a big deal, just curious.

test/Microsoft.AspNet.Localization.FunctionalTests/LocalizationTest.cs
```diff
@@ -30,6 +30,26 @@ public class LocalizationTest
...
+            RuntimeArchitecture runtimeArchitechture)
+        {
+            var testRunner = new TestRunner();
+            return testRunner.RunTestAndVerifyResponse(
+                runtimeFlavor,
+                runtimeArchitechture,
+                applicationBaseUrl,
+                "ResourcesInFolder",
+                "fr-FR-test",
+                "Bonjour from StartupResourcesInFolder Bonjour from Test in resources folder Bonjour from Customer in resources folder");
```
ryanbrandenburg: Like @hishamco said, you'll have to modify these to fit with the changes from my PR.

src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
```diff
@@ -142,6 +154,39 @@ private CultureInfo GetCultureInfo(IList<string> cultures, IList<CultureInfo> su
...
+                    else
+                    {
+                        // Culture name was null so just remove it
+                        cultureNames.RemoveAt(i);
+                    }
+                }
+
+                if (cultureNames.Count > 0)
+                {
+                    return GetCultureInfo(cultureNames, supportedCultures, fallbackToAncestorCulture);
```
ryanbrandenburg: So this would recurse through ["fr-FR-CA"] => ["fr-FR"] => ["fr"] trying to get the culture info for each one each time? Makes sense.

ryanbrandenburg:
I vote :shipit: once you fix the new tests to match my changes.

src/Microsoft.AspNet.Localization/RequestLocalizationOptions.cs
DamianEdwards: Yeah an ancestor is at any level above me, parent is the direct level.

src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
```diff
@@ -142,6 +156,39 @@ private CultureInfo GetCultureInfo(IList<string> cultures, IList<CultureInfo> su
...
+            if (fallbackToAncestorCulture & currentDepth < MaxCultureFallbackDepth)
+            {
+                // Walk backwards through the culture list and remove any root cultures (those with no parent)
+                for (var i = cultureNames.Count - 1; i >= 0; i--)
+                {
+                    var cultureName = cultureNames[i];
+                    if (cultureName != null)
+                    {
+                        var lastIndexOfHyphen = cultureName.LastIndexOf('-');
+                        if (lastIndexOfHyphen > 0)
```
rynowak: Why is it that we need to parse manually instead of relying on `CultureInfo.Parent`?

rynowak: is `-bleh` a valid name?

src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
```diff
@@ -142,6 +156,39 @@ private CultureInfo GetCultureInfo(IList<string> cultures, IList<CultureInfo> su
...
+                        {
+                            // The culture had no sections left to trim so remove it from the list of candidates
+                            cultureNames.RemoveAt(i);
+                        }
+                    }
+                    else
+                    {
+                        // Culture name was null so just remove it
+                        cultureNames.RemoveAt(i);
+                    }
```
rynowak: Could collapse a few cases here and simplify the loop like:
```
var cultureName = cultureNames[i];
cultureNames.RemoveAt(i);

int index;
if (cultureName != null && (index = cultureName.LastIndexOf("-")) >= 0)
{
     cultureNames.Insert(i, cultureName.Substring(0, index);
}
```

Your call if documenting the negative cases is valuable, this is reasonable either way.

rynowak:
:shipit: from me

src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
DamianEdwards: The point is to avoid creating instances of `CultureInfo` at all because it's expensive.

We're not overly concerned on what is strictly valid in this case, as ultimately it's checked against the list of cultures the application explicitly stated it supported.

src/Microsoft.AspNet.Localization/RequestLocalizationOptions.cs
Eilon: The verb is `fall back`, so this should be `FallBackToAncestorCulture`. But I don't super duper care.

miloush: @DamianEdwards ancestor and parent words might have these semantics, but I have to agree with @ryanbrandenburg that there is no mention about ancestor cultures anywhere (and it also has a temporal connotation), while parent cultures are well known concept to both .NET and native developers

DamianEdwards: Yeah fair enough (on both counts). I'll change it to `FallBackToParentCultures`

src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
miloush: @rynowak no, leading and trailing hyphens are not valid, more naming rules at MSDN [CultureAndRegionInfoBuilder](https://msdn.microsoft.com/en-us/library/system.globalization.cultureandregioninfobuilder.cultureandregioninfobuilder.aspx) (which does not mean you cannot get it from a client of course)

@DamianEdwards  For the record, there is a fundamental difference between parsing and `CultureInfo.Parent`. There are parent chains like:
`zh-Hans-HK -> zh-CHS -> zh-Hans -> zh -> Invariant` or 
`zh-TW -> zh-CHT -> zh-Hant -> zh -> Invariant` or 
`mn-MN -> mn-Cyrl -> mn -> Invariant` etc.
Supporting simplified and traditional Chinese while expecting all of `zh-CN`, `zh-HK`, `zh-MO`, `zh-SG`, `zh-TW`, `zh-CHS` and `zh-CHT` to resolve to the correct one sounds reasonable (and common) to me. I am also sure there was a political review for these decisions. Anyway, someone from the team should probably decide about these costs vs performance.

miloush:
Just curious about the breadth first approach. Do I understand it correctly that if
```
en-US
de
```
is requested and
```
en
de
```
are available, then `de` would be preferred? Is that due to any backward compatibility or simply performance? Or do we just assume that this does not usually happen since clients automatically add neutral culture before others?

DamianEdwards:
@miloush after some discussion we decided depth-first is probably more useful here, hence #177

Issue #167, opened 2015-12-10
ResourceManagerStringLocalizerFactory should look for resource name without the application name prefix on the type name when Resources path is set


DamianEdwards:
When `LocalizationOptions.ResourcesPath` is set, make `ResourceManagerStringLocalizerFactory` use a resource location without the application name prefix.

e.g. where `location` is `MyApp.Controllers.HomeController` and `LocalizationOptions.ResourcesPath` is "Resources", `ResourceManagerStringLocalizerFactory` would look for a resource with base name "MyApp.Resources.Controllers.HomeController".

PR #176, opened 2015-12-16
Damianedwards/app prefix

DamianEdwards:
`ResourceManagerStringLocalizerFactory` now trims the application name from the start of the type name/location when `LocalizationOptions.ResourcesPath` is set, so that `IStringLocalizer<MyController>` will look for resource name `MyApplication.Resources.Controllers.MyController` instead of `MyApplication.Resources.MyApplication.Controllers.MyController`

#167

@rynowak @ryanbrandenburg


src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -69,7 +68,8 @@ public IStringLocalizer Create(Type resourceSource)
 
             var baseName = string.IsNullOrEmpty(_resourcesRelativePath)
                 ? typeInfo.FullName
-                : _applicationEnvironment.ApplicationName + "." + _resourcesRelativePath + typeInfo.FullName;
+                : _applicationEnvironment.ApplicationName + "." + _resourcesRelativePath
+                    + TrimPrefix(typeInfo.FullName, _applicationEnvironment.ApplicationName + ".");
```
rynowak: This is some ugly code

DamianEdwards: I don't disagree.

ryanbrandenburg:
Seems fine. :shipit:

PR #177, opened 2015-12-16
Change culture fallback to depth-first

DamianEdwards:
After talking to some folks we think depth-first is a better approach here (i.e. more useful).

@lodejard @ryanbrandenburg @rynowak @miloush

#112

### DIFF
--- a/src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
+++ b/src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
@@ -78,8 +78,7 @@ namespace Microsoft.AspNet.Localization
                             cultureInfo = GetCultureInfo(
                                 cultures,
                                 _options.SupportedCultures,
-                                _options.FallBackToParentCultures,
-                                currentDepth: 0);
+                                _options.FallBackToParentCultures);
                         }
 
                         if (_options.SupportedUICultures != null)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Localization/RequestLocalizationMiddleware.cs
```diff
@@ -156,40 +153,31 @@ private static void SetCurrentThreadCulture(RequestCulture requestCulture)
...
+                var lastIndexOfHyphen = cultureName.LastIndexOf('-');
 
-                if (cultureNames.Count > 0)
+                if (lastIndexOfHyphen > 0)
                 {
-                    return GetCultureInfo(cultureNames, supportedCultures, fallbackToAncestorCulture, currentDepth + 1);
+                    // Trim the trailing section from the culture name, e.g. "fr-FR" becomes "fr"
+                    var parentCultureName = cultureName.Substring(0, lastIndexOfHyphen);
+
+                    culture = GetCultureInfo(parentCultureName, supportedCultures, fallbackToAncestorCulture, currentDepth + 1);
```
ryanbrandenburg: Definitely a lot cleaner this way.

ryanbrandenburg:
This looks pretty good, the only thing I might say is that we could note on the Fallback option doc-comment that this is "best effort" and won't ALWAYS give you every possible parent. Otherwise I'd say :shipit:

PR #181, opened 2015-12-18
Tweaked cache key when getting localizer from basName/location:

DamianEdwards:
- Cleaned up code a little to make it easier to understand (maybe)
- #138 

@ryanbrandenburg  @rynowak


rynowak:
:shipit: - seems simple enough. Should we have a test that we can retrieve the same one twice (with/without the resource name set)?

DamianEdwards:
Sounds like a good ide, will add a test.

ryanbrandenburg:
:shipit: simple cleanup it seems to me.

PR #184, opened 2015-12-21
Throw proper exception when missing resource

### LINKED ISSUE #143: GetAllStrings give System.ArgumentNullException
GetAllStrings give `System.ArgumentNullException`.because `cultureResourceStream` is null

The problem seems to be in `private IList<string> GetResourceNamesForCulture(CultureInfo culture)` in `ResourceManagerStringLocalizer.cs`

 ```
var cultureResourceNames = _resourceNamesCache.GetOrAdd(cacheKey, _ =>
            {
                var names = new List<string>();
                using (var cultureResourceStream = _resourceAssemblyWrapper.GetManifestResourceStream(resourceStreamName))
                using (var resources = new ResourceReader(cultureResourceStream))
                {
                    foreach (DictionaryEntry entry in resources)
                    {
                        var resourceName = (string)entry.Key;
                        names.Add(resourceName);
                    }
                }
                return names;
            });
```

If have 3 files but only Invariant can be read (`HomeController.nl-NL.resx`,`HomeController.nl.resx`, `HomeController.resx`)
The files are ok and can be read by `resourceManager.GetString(name)` command

### PR DESCRIPTION
**Pull request by ryanbrandenburg**
Fixes #143

### CODE REVIEW

**Comment by ryanbrandenburg** [comment]
@NTaylorMullen @DamianEdwards if you guys could review me here?

File: src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs [hunk 1]
```diff
@@ -236,12 +236,21 @@ private IList<string> GetResourceNamesForCulture(CultureInfo culture)
...
+                            foreach (DictionaryEntry entry in resources)
+                            {
+                                var resourceName = (string)entry.Key;
+                                names.Add(resourceName);
+                            }
+                        }
+                    }
+                    else
+                    {
+                        throw new MissingManifestResourceException();
```
NTaylorMullen: Should provide a message. The default one is `Unable to find manifest resource.` Would make it incredibly difficult for a user to decipher why something is failing.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -236,12 +236,21 @@ private IList<string> GetResourceNamesForCulture(CultureInfo culture)
             {
                 var names = new List<string>();
                 using (var cultureResourceStream = _resourceAssemblyWrapper.GetManifestResourceStream(resourceStreamName))
-                using (var resources = new ResourceReader(cultureResourceStream))
                 {
-                    foreach (DictionaryEntry entry in resources)
+                    if (cultureResourceStream != null)
```
pranavkm: Consider inverting the condition so you wouldn't have to nest the body:

```C#
if (cultureResourceStream  == null)
{
   throw new MissingManifestResourceException();
}

using (var resources = new ResourceReader(cultureResourceStream))
{

}
```

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -59,10 +59,33 @@ public void EnumeratorCacheIsScopedByAssembly()
             Assert.Equal(expectedCallCount, resourceAssembly2.GetManifestResourceStreamCallCount);
         }
 
+        [Fact]
+        public void ResourceManagerStringLocalizer_GetAllStrings_MissingResourceDoesntThrow()
```
NTaylorMullen: Should have a throw test-case. Xunit is awesome with them. `Assert.Throws` is your friend :smile:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -236,12 +236,21 @@ private IList<string> GetResourceNamesForCulture(CultureInfo culture)
             {
                 var names = new List<string>();
```
pranavkm: You could declare this right above the `foreach` and return it from there. Does't seem like it's used outside of the loop.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -236,12 +236,21 @@ private IList<string> GetResourceNamesForCulture(CultureInfo culture)
...
-                using (var resources = new ResourceReader(cultureResourceStream))
                 {
-                    foreach (DictionaryEntry entry in resources)
+                    if (cultureResourceStream != null)
                     {
-                        var resourceName = (string)entry.Key;
-                        names.Add(resourceName);
+                        using (var resources = new ResourceReader(cultureResourceStream))
+                        {
+                            foreach (DictionaryEntry entry in resources)
```
pranavkm: `var`?

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -59,10 +59,33 @@ public void EnumeratorCacheIsScopedByAssembly()
...
+            var resourceManager = new TestResourceManager(baseName, resourceAssembly.Assembly);
+            var localizer = new ResourceManagerWithCultureStringLocalizer(
+                resourceManager,
+                resourceAssembly.Assembly,
+                baseName,
+                resourceNamesCache,
+                CultureInfo.CurrentCulture);
+
+            // Act
+            var t = localizer.GetAllStrings(true);
```
pranavkm: `var result`

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
pranavkm: Could this include the resource name you were looking for? Would help in diagnosing the issue.

pranavkm:
:watch:

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
NTaylorMullen: Somewhat odd that passing `true` here supresses thrown exceptions where passing `false` would cause it to throw

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
NTaylorMullen: Can't. It'd be of type `object`.

NTaylorMullen:
:watch:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: I had though the only place this was called would eat and exceptions but its also in GetAllStrings, so I'll do this.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: A using block counts as a different scope so it would be inaccessible to the return if i put it in there, so I have to move the return up into it too,

pranavkm: Wouldn't that be ok?

ryanbrandenburg: Oh yes, I had already made the change, was just noting that that wasn't the only change there.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
ryanbrandenburg: Yeah I agree. Would the proper fix there be to eat the missing manifest exception in all cases?

NTaylorMullen: Why does it supress the exception in the `true` case? Seems odd to hide that info from the user, i'm probably missing something.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
ryanbrandenburg: If we do catch the excetions like I suggested below then there will be no throw case (though I should add a case for not following parents.)

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
ryanbrandenburg: Because the true case means walking up the list of parents and looking for each of those resources, the failure of which to exist isn't really exceptional.

NTaylorMullen: If you walk up all the parents and never find one I'd expect to see an exception. That way the user can fix their code.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -207,7 +210,11 @@ private IEnumerable<string> GetResourceNamesFromCultureHierarchy(CultureInfo sta
                         resourceNames.Add(resourceName);
                     }
                 }
-                catch (MissingManifestResourceException) { }
+                catch (MissingManifestResourceException)
+                {
+                    missingCultures++;
```
pranavkm: Wouldn't the two numbers be different as soon as one of the parent throws?

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -59,10 +59,35 @@ public void EnumeratorCacheIsScopedByAssembly()
...
+                resourceManager,
+                resourceAssembly.Assembly,
+                baseName,
+                resourceNamesCache,
+                CultureInfo.CurrentCulture);
+
+            // Act & Assert
+            Assert.Throws<MissingManifestResourceException>(() =>
+            {
+                //We have to access the result so it evaluates.
```
pranavkm: Space after the comment. Also `ToArray` instead of `Count` (just because that's what we typically use to evaluate `IEnumerable`s elsewhere.

ryanbrandenburg:
:up: :date:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: Yes, but the only case it should throw under is when all cultures are missing (missingCultures == walked cultutres) It's kinda a weird way to go about it but I don't know of a way to do an effective "not caught")

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -207,7 +210,11 @@ private IEnumerable<string> GetResourceNamesFromCultureHierarchy(CultureInfo sta
                         resourceNames.Add(resourceName);
                     }
                 }
-                catch (MissingManifestResourceException) { }
+                catch (MissingManifestResourceException)
```
NTaylorMullen: In `GetResourceNamesForCulture` you're manually throwing and then catching here. Is there another call site that's throwing `MissingManifestResourceException` or could you just return false and flow that response up? Try -> catching like this can be very expensive and inneficient given what you do below :smile:

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -59,10 +59,35 @@ public void EnumeratorCacheIsScopedByAssembly()
...
+            var resourceManager = new TestResourceManager(baseName, resourceAssembly.Assembly);
+            var localizer = new ResourceManagerWithCultureStringLocalizer(
+                resourceManager,
+                resourceAssembly.Assembly,
+                baseName,
+                resourceNamesCache,
+                CultureInfo.CurrentCulture);
+
+            // Act & Assert
+            Assert.Throws<MissingManifestResourceException>(() =>
```
NTaylorMullen: Assert the message so we know it's the custom exception you're throwing. Can do:

```C#
var exception = Assert.Throws<MissingManifestResourceException>(() => ...
Assert.Equals("The exception message.", exception.Message, StringComparer.Ordinal);
```

NTaylorMullen:
:watch:

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
pranavkm: `Assert.Equal` should perfom case sensitive comparison. You don't need to specify the comparer.

pranavkm:
:shipit:

NTaylorMullen:
Looks like a change hid this. Re-raising it: https://github.com/aspnet/Localization/pull/184#discussion_r48516258

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: Let me know if my change is what you were asking for.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -197,17 +204,21 @@ private IEnumerable<string> GetResourceNamesFromCultureHierarchy(CultureInfo sta
             var currentCulture = startingCulture;
             var resourceNames = new HashSet<string>();
 
+            var hasAnyCultures = false;
```
NTaylorMullen: Cant this also be derived by the size of `resourceNames`? Aka if it has at least 1 element

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -153,6 +153,13 @@ protected IEnumerable<LocalizedString> GetAllStrings(bool includeParentCultures,
                 ? GetResourceNamesFromCultureHierarchy(culture)
                 : GetResourceNamesForCulture(culture);
 
+            if (resourceNames == null && !includeParentCultures)
+            {
+                var resourceStreamName = GetResourceStreamName(culture);
+                throw new MissingManifestResourceException(
+                            string.Format(Resources.Localization_MissingManifest, resourceStreamName));
```
NTaylorMullen: @pranavkm did we do away with the `Resources.FormatXYZ` auto-generated code?

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -59,10 +59,59 @@ public void EnumeratorCacheIsScopedByAssembly()
             Assert.Equal(expectedCallCount, resourceAssembly2.GetManifestResourceStreamCallCount);
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void ResourceManagerStringLocalizer_GetAllStrings_ExistingResourceDoesntThrow(bool includeParentCultures)
+        {
+            // Arrange var resourceNamesCache = new ResourceNamesCache();
```
NTaylorMullen: `// Arrange`

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: I don't think we strictly can, as the resource names are the names of the KVP's, and if the resx file were empty but existent then we would say they were missing a manifest, which would be wrong.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -59,10 +59,59 @@ public void EnumeratorCacheIsScopedByAssembly()
...
+            var resourceAssembly = new TestAssemblyWrapper();
+            var resourceManager = new TestResourceManager(baseName, resourceAssembly.Assembly);
+            var localizer = new ResourceManagerStringLocalizer(resourceManager, resourceAssembly, baseName, resourceNamesCache);
+
+            // Act
+            // We have to access the result so it evaluates.
+            var strings = localizer.GetAllStrings(includeParentCultures).ToList();
+
+            // Assert
+            Assert.Equal(1, strings.Count());
```
NTaylorMullen:
```C#
var value = Assert.Single(strings);
Assert.Equal("thevalue", value)
```

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: Yeah I looked for that too but it didn't show as an option.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -59,10 +59,59 @@ public void EnumeratorCacheIsScopedByAssembly()
             Assert.Equal(expectedCallCount, resourceAssembly2.GetManifestResourceStreamCallCount);
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void ResourceManagerStringLocalizer_GetAllStrings_ExistingResourceDoesntThrow(bool includeParentCultures)
```
NTaylorMullen: `ResourceManagerStringLocalizer_GetAllStrings_ReturnsExpectedValues`

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
NTaylorMullen: @pranavkm was talking about eventually doing away with it. What happens if you run `./build resx` at the root of the repo?

pranavkm: Nope. Run `build resx` and it should be generated.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
NTaylorMullen: Ok fair :+1:

NTaylorMullen:
:watch:

ryanbrandenburg:
:up: :date:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -218,10 +229,15 @@ private IEnumerable<string> GetResourceNamesFromCultureHierarchy(CultureInfo sta
                 currentCulture = currentCulture.Parent;
             }
 
+            if (!hasAnyCultures)
+            {
+                throw new MissingManifestResourceException(Resources.FormatLocalization_MissingManifest_Parent());
```
NTaylorMullen: Only use Format when there are parameters to the resource string. Can just do: `Resources.Localization_MissingManifest_Parent`

NTaylorMullen:
:shipit:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -153,6 +153,13 @@ protected IEnumerable<LocalizedString> GetAllStrings(bool includeParentCultures,
                 ? GetResourceNamesFromCultureHierarchy(culture)
                 : GetResourceNamesForCulture(culture);
 
+            if (resourceNames == null && !includeParentCultures)
+            {
+                var resourceStreamName = GetResourceStreamName(culture);
+                throw new MissingManifestResourceException(
+                            Resources.FormatLocalization_MissingManifest(resourceStreamName));
```
pranavkm: Weird indentation.

PR #194, opened 2016-01-02
Adding TimeZoneInfo to the RequestCulture class

steentottrup:
I don't expect this PR to be accepted, but hope it will start a talk about how we get the TimeZone information into the request somehow.

I understand this can not be solved in aspnet5, but any change needs to be put into the main .NET framework too.

It would be nice with a simpler solution to handle time zones properly in web applications, in the generated output. When storing dates in databases as UTC, and with authenticated users where you know their time zone, it would be sweet if ToShortTime and other methods used for outputting times would "just work", like they do for dates and the threads CultureInfo setting.

### DIFF
--- a/samples/LocalizationSample/Startup.cs
+++ b/samples/LocalizationSample/Startup.cs
@@ -25,7 +25,7 @@ namespace LocalizationSample
         {
             app.UseRequestLocalization(options =>
             {
-                options.DefaultRequestCulture = new RequestCulture("en-US");
+                options.DefaultRequestCulture = new RequestCulture("en-US", TimeZoneInfo.Local.Id);
 
                 // Set options here to change middleware behavior
                 options.SupportedCultures = new List<CultureInfo>
--- a/src/Microsoft.AspNet.Localization/RequestCulture.cs
+++ b/src/Microsoft.AspNet.Localization/RequestCulture.cs
@@ -16,8 +16,9 @@ namespace Microsoft.AspNet.Localization
         /// properties set to the same <see cref="CultureInfo"/> value.
         /// </summary>
         /// <param name="culture">The <see cref="CultureInfo"/> for the request.</param>
-        public RequestCulture(CultureInfo culture)
-            : this(culture, culture)
+		/// <param name="timeZone">The <see cref="TimeZoneInfo"/> for the request.</param>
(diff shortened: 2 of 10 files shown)


davidfowl:
What about showing more of an end to end example? This doesn't look like much without showing the consumption side.

steentottrup:
Let me start off by giving you an example in writing, and then I'll try getting an end to end example into the repo later on.

I've been working with content management system for a long time, and several of my clients have people working on their web site content from several time zones. This always poses a problem, as the server is placed in one time zone, and the users are across several. So when you want to publish some page at 9:00 AM (your time zone), what time do you actually write as publishing time?

So you ask the user to select a language (cultureinfo) and a time zone as part of the properties on their account. If you've ever used some "old school" forum system, like phpBB you know what I'm talking about (or a more recent one, like MVC Forum).

So .NET (and ASP.NET) has always had CultureInfo as part of a thread, so we can expect dates to be formatted properly and in the correct language (as long as I can remember anyway). This is done automatically in all the "format" methods on the DateTime object (ToShortDateString(), ToLongDateString() etc.), by using the CultureInfo set on the current thread. The year is even correct when the calendar in use uses another year than the "normal" one.

But what about the time part of a DateTime? We have the DateTimeKind property, but only 3 options, Utc, Local (as in server-time) or Unspecified. No matter what I select for DateTimeKind for my DateTime object, it will be written as the same time if I use ToString().

So when I go fetch my DateTime value from the database, where I usually store it as UTC, I would like to be able to use ToShortTimeString and it "just works". The time will be show correct, from the user's viewpoint. So when a page is published at 9 PM his time zone, that is what the UI will show, and if another user with a different time zone viewed the same data, the time would be different, but still correct (this actually involves both TimeZoneInfo and CultureInfo, because of daylight saving...).

From time to time I've "hacked" this by creating my own CultureInfo implementation, where the only difference is that it includes a TimeZoneInfo property. That way I can store it on the thread, and by using an extension method for showing time, fetch it there, like all the DateTime format method do anyway. It just didn't feel quite right.

So if this was "baked" into .NET, and ASP.NET, it would be easier, and "just work".

hishamco:
It will be nice if this opened as discussion before submitting the PR, further more CI fails
thanks @steentottrup for the point that you mentioned

steentottrup:
So you suggest I kill this PR and add an issue instead?

PR #195, opened 2016-01-05
 * Make test culture independent

ryanbrandenburg:
Should fix a problem we were having on CI where the culture was "en_POSIX" instead of "en_US".

### DIFF
--- a/test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
+++ b/test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Extensions.Localization.Tests
             });
             var expected = includeParentCultures
                 ? "No manifests exist for the current culture."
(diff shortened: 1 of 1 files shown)


test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerTest.cs
```diff
@@ -105,7 +105,7 @@ public void ResourceManagerStringLocalizer_GetAllStrings_MissingResourceThrows(b
             });
             var expected = includeParentCultures
                 ? "No manifests exist for the current culture."
-                : "The manifest 'testington.en-US.resources' was not found.";
+                : $"The manifest 'testington.{CultureInfo.CurrentCulture.ToString()}.resources' was not found.";
```
pranavkm: Shouldn't need the `ToString()`
If it's important to keep `en-us` you could use `ReplaceCulture` (but this looks fine as-is).

pranavkm:
:shipit:

Issue #199, opened 2016-01-13
Switch from System.Resources.ReaderWriter package to new separate Reader and Writer packages


Eilon:
The System.Resources.ReaderWriter was split into two packages: System.Resources.Reader and System.Resources.Writer.

We need to switch to use the new packages (or just one of them, if we need only one). It's currently in use here:
* https://github.com/aspnet/Localization/blob/489250a8e1c97d2b415b1c8994f1943b6d229d73/src/Microsoft.Extensions.Localization/project.json#L23

This appears to be the only real usage of this assembly. (It's also used in DNX, but that won't be an issue in the near future.)

Separately, we'll have to update the [NuGetPackageVerifier list](https://github.com/aspnet/Coherence-Signed/blob/c5c5ce686e33a5d683e36a8ea15217e9ecb339b9/tools/NugetPackageVerifier.json#L938) to change the metadata for this, but we'll presumably see that failure once we accept the new bits.

hishamco:
I think this should be closed after merge

ryanbrandenburg:
Ah, thanks for the reminder @hishamco

PR #200, opened 2016-01-14
 * Replace System.Resources.ReaderWriter

ryanbrandenburg:
Part of #199.

### DIFF
--- a/src/Microsoft.Extensions.Localization/project.json
+++ b/src/Microsoft.Extensions.Localization/project.json
@@ -20,7 +20,7 @@
     "dotnet5.4": {
       "dependencies": {
         "System.Collections.Concurrent": "4.0.11-*",
-        "System.Resources.ReaderWriter": "4.0.0-*"
+        "System.Resources.Reader": "4.0.0-*"
       }
     }
   }
--- a/test/Microsoft.Extensions.Localization.Tests/project.json
+++ b/test/Microsoft.Extensions.Localization.Tests/project.json
@@ -17,7 +17,8 @@
     },
     "dnxcore50": {
       "dependencies": {
-        "moq.netcore": "4.4.0-beta8"
+        "moq.netcore": "4.4.0-beta8",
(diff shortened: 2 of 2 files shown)


src/Microsoft.Extensions.Localization/project.json
```diff
@@ -16,11 +16,12 @@
...
   },
   "frameworks": {
-    "net451": {},
+    "net451": { },
     "dotnet5.4": {
       "dependencies": {
         "System.Collections.Concurrent": "4.0.11-*",
-        "System.Resources.ReaderWriter": "4.0.0-*"
+        "System.Resources.Reader": "4.0.0-*",
+        "System.Resources.Writer": "4.0.0-*"
```
pranavkm: Why do we need the writer here?

Eilon:
:shipit:

BrennanConroy:
Don't push until we grab newer CoreCLR bits since this uses a new package we don't have on our feed yet.

Eilon:
@BrennanConroy yup, aware of that :smile:

ryanbrandenburg:
@BrennanConroy ok, let me know when you guys have that done.

KrzysztofCwalina:
BTW, we recently changed the feeds. This is where the component is located: https://www.myget.org/feed/dotnet-core/package/nuget/System.Resources.Reader

Eilon:
@KrzysztofCwalina we don't pull directly from those feeds in this build - we always take snapshots of CoreCLR as part of a separate process. However, isn't that the same feed as it always was, though?

BrennanConroy:
Ya, thats the feed we've always used

KrzysztofCwalina:
We changed from dotnet-corefx to dotnet-core feed. I think the reson is that we ran out of space on the old feed ;-(

Eilon:
@KrzysztofCwalina ah ok I don't see us using the old feed anywhere. I'm pretty sure we always used the `dotnet-core` feed (not `dotnet-corefx`).

PR #202, opened 2016-01-19
Change build to use dotnet

### LINKED ISSUE #198: Change to build to use dotnet

pranavkm:
Fixes #198

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -48,3 +48,5 @@
 *.fsproj text=auto
 *.dbproj text=auto
 *.sln text=auto eol=crlf
+
+*.sh eol=lf
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 [Bb]in/
 TestResults/
 .nuget/
+.build/
+.testPublish/
 *.sln.ide/
 _ReSharper.*/
 packages/
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ addons:
     - libssl-dev
     - libunwind8
     - zlib1g
-env:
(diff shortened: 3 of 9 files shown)


pranavkm:
cc @ryanbrandenburg \ @victorhurdugaci

victorhurdugaci:
the change looks okay but you have some test failures

.gitattributes
```diff
@@ -48,3 +48,5 @@
 *.fsproj text=auto
 *.dbproj text=auto
 *.sln text=auto eol=crlf
+
+*.sh eol=lf
```
ryanbrandenburg: I assume this was intentional, why do we define the EOL for .sh and not for most others?

victorhurdugaci: sh files cannot be executed if the eol is CRLF

ryanbrandenburg:
Yeah, seems like the application path is either not being set to the correct place or not being created.

.gitattributes
ryanbrandenburg: Huh, learn something new every day.

pranavkm:
Nice catch @victorhurdugaci. The path seems incorrect although it seemed to pass locally. Will update

NuGet.config
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetcidev/api/v3/index.json" />
+    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvolatiledev/api/v3/index.json" />
```
pranavkm: Will undo this before checking it in.

victorhurdugaci:
:+1:

PR #213, opened 2016-02-17
Localization Sample Fix

hishamco:
After option pattern has been applied to localization repo, seems you can't use ```RequestCultureProviders.Insert``` inside ```UseRequestLocalization``` because ```RequestCultureProviders``` will get its value during the instantiation process. While the [line](https://github.com/aspnet/Localization/blob/dev/samples/LocalizationSample/Startup.cs#L49) is commented perhaps no one notice that you can't invoke ```Insert``` method anymore
/cc @Eilon

### DIFF
--- a/samples/LocalizationSample/Startup.cs
+++ b/samples/LocalizationSample/Startup.cs
@@ -36,21 +36,19 @@ namespace LocalizationSample
 #if !DNXCORE50
             supportedCultures.Add(new CultureInfo("zh-CHT"));
 #endif
-            app.UseRequestLocalization(new RequestLocalizationOptions
-            {
+            var options = new RequestLocalizationOptions {
                 DefaultRequestCulture = new RequestCulture("en-US"),
-
-                // Set options here to change middleware behavior
                 SupportedCultures = supportedCultures,
                 SupportedUICultures = supportedCultures
+            };
+            // Optionally create an app-specific provider with just a delegate, e.g. look up user preference from DB.
+            // Inserting it as position 0 ensures it has priority over any of the default providers.
+            //options.RequestCultureProviders.Insert(0, new CustomRequestCultureProvider(async context =>
+            //{
 
-                // Optionally create an app-specific provider with just a delegate, e.g. look up user preference from DB.
-                // Inserting it as position 0 ensures it has priority over any of the default providers.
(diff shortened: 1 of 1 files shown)


Eilon:
@ryanbrandenburg can you have a look?

ryanbrandenburg:
It seems like this is more a matter of preference about what we want to showcase here, because I don't think we can really show both. We could do what this PR does, showcasing the ability to fiddle with the CultureProviders (but only in a comment, and not one that would even work if you just un-commented it), or we could showcase the new options style like @JunTaoLuo had done before this. I would say that if we decide it's more important to us to showcase the options we should remove that comment as it's misleading.
CC @JunTaoLuo and @Eilon for opinions.

hishamco:
Can you try to uncomment the [line](https://github.com/aspnet/Localization/blob/dev/samples/LocalizationSample/Startup.cs#L49) and check whether you can call the ```Insert``` method or not

Eilon:
Uncommenting the line in the original code certainly won't work - object initializers can't invoke methods - only properties (and maybe fields) can be set.

I think this change is actually alright. With this change it's a bit more verbose, but you can uncomment the code and it should work. The other option is to just remove all the commented code.

hishamco:
Exactly @Eilon but I comment the line to leave the ```RequestCultureProviders``` sequence as it is, otherwise it will be the winner provider which is not preferred in this case unless you uncomment the line

JunTaoLuo:
I agree, this change illustrate the correct way of configuring the options. The only comment I have is that the `CustomRequestCultrureProvider` can't be created as is after uncommenting the code. It would be nice to add some dummy `Task` that returns an empty `ProviderCultrueResult` just to illustrate what is required to actually create an app-specific provider. However, since we never did that previously, it isn't a big deal if we just leave it the way it is.

ryanbrandenburg:
Sounds like we're going with this one then, seems reasonable.

ryanbrandenburg:
Merged in a9289820552

hishamco:
Shall I apply the same thing in other places such as Entopy repo?

Eilon:
@hishamco you mean loc sample in Entropy? Sure.

hishamco:
thanks

PR #216, opened 2016-02-19
Enabled xml doc generation

ajaybhargavb:
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Localization/IRequestCultureFeature.cs
+++ b/src/Microsoft.AspNetCore.Localization/IRequestCultureFeature.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Localization
         /// <summary>
         /// The <see cref="IRequestCultureProvider"/> that determined the request's culture information.
         /// If the value is <c>null</c> then no provider was used and the request's culture was set to the value of
-        /// <see cref="RequestLocalizationOptions.DefaultRequestCulture"/>.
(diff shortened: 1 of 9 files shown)


NTaylorMullen:
:shipit:

src/Microsoft.AspNetCore.Localization/IRequestCultureFeature.cs
```diff
@@ -16,7 +16,7 @@ public interface IRequestCultureFeature
         /// <summary>
         /// The <see cref="IRequestCultureProvider"/> that determined the request's culture information.
         /// If the value is <c>null</c> then no provider was used and the request's culture was set to the value of
-        /// <see cref="RequestLocalizationOptions.DefaultRequestCulture"/>.
+        /// <see cref="Builder.RequestLocalizationOptions.DefaultRequestCulture"/>.
```
hishamco: @ajaybhargavb why you are adding **Builder**?!! Is the xml doc generation will discover that after building process or there's something else

pranavkm: `RequestLocalizationOptions` is located under `Microsoft.AspNetCore.Builder` which isn't referenced here.

hishamco: Oops, you know what this the first time I notice that, I thought it's under ```Microsoft.AspNetCore.Localization``` , It's my fault .. thanks

PR #219, opened 2016-02-25
Don't run functional tests in parallel

pranavkm:


pranavkm:
cc @dougbu

dougbu:
:shipit: once at least one of each platform CI completes.

JunTaoLuo:
I thought we add these settings on the CI?

pranavkm:
@JunTaoLuo we run into the same issues building locally too.

dougbu:
> I thought we add these settings on the CI?

How does that help Travis builds or local testing?

JunTaoLuo:
Will this override the CI's setting of the env var `NO_PARALLEL_TEST_PROJECTS` after running Localization? We might need to add this setting to all the repos that run functional tests then.

dougbu:
> Will this override the CI's setting of the env var NO_PARALLEL_TEST_PROJECTS

This change is local to the process executing the Localization run. Will not affect the process executing `.shade` files in Universe.

> We might need to add this setting to all the repos that run functional tests then.

Here the issue was multiple tests attempting to start server instances and hitting file contention. So, not all functional tests. For example functional tests using an in-memory server shouldn't (normally) need anything similar.

JunTaoLuo:
Ah I see, I misunderstood a few things.

PR #220, opened 2016-03-01
Fix tests after adding Options service #217

JunTaoLuo:
@javiercn

### DIFF
--- a/test/Microsoft.Extensions.Localization.Tests/LocalizationServiceCollectionExtensionsTest.cs
+++ b/test/Microsoft.Extensions.Localization.Tests/LocalizationServiceCollectionExtensionsTest.cs
@@ -21,15 +21,33 @@ namespace Microsoft.Extensions.Localization.Test
 
             // Assert
             var services = collection.ToList();
-            Assert.Equal(2, services.Count);
+            Assert.Equal(4, services.Count);
 
-            Assert.Equal(typeof(IStringLocalizerFactory), services[0].ServiceType);
-            Assert.Equal(typeof(ResourceManagerStringLocalizerFactory), services[0].ImplementationType);
-            Assert.Equal(ServiceLifetime.Singleton, services[0].Lifetime);
-
-            Assert.Equal(typeof(IStringLocalizer<>), services[1].ServiceType);
-            Assert.Equal(typeof(StringLocalizer<>), services[1].ImplementationType);
-            Assert.Equal(ServiceLifetime.Transient, services[1].Lifetime);
+            Assert.Collection(services,
+                service =>
+                {
+                    Assert.Equal(typeof(IOptions<>), service.ServiceType);
+                    Assert.Equal(typeof(OptionsManager<>), service.ImplementationType);
+                    Assert.Equal(ServiceLifetime.Singleton, service.Lifetime);
+                },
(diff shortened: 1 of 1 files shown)


test/Microsoft.Extensions.Localization.Tests/LocalizationServiceCollectionExtensionsTest.cs
```diff
@@ -21,15 +21,23 @@ public void AddLocalization_AddsNeededServices()
 
             // Assert
             var services = collection.ToList();
-            Assert.Equal(2, services.Count);
+            Assert.Equal(4, services.Count);
```
pranavkm:
```C#
Assert.Collection(services,
    service => 
    {
        Assert.Equal(typeof(IOptions<>), service.ServiceType);
        Assert.Equal(typeof(OptionsManager<>), service.ImplementationType);
        Assert.Equal(ServiceLifetime.Singleton, service.Lifetime);
    },
    service =>
    {
        Assert.Equal(typeof(IOptionsMonitor<>), service.ServiceType);
        Assert.Equal(typeof(OptionsMonitor<>), service.ImplementationType);
        Assert.Equal(ServiceLifetime.Singleton, service.Lifetime);
    },
    service =>
    {
        Assert.Equal(typeof(IStringLocalizerFactory), service.ServiceType);
        Assert.Equal(typeof(ResourceManagerStringLocalizerFactory), service.ImplementationType);
        Assert.Equal(ServiceLifetime.Singleton, service.Lifetime);
    },
    service =>
    {
        Assert.Equal(typeof(IStringLocalizer<>), service.ServiceType);
            Assert.Equal(typeof(StringLocalizer<>), service.ImplementationType);
            Assert.Equal(ServiceLifetime.Transient, service.Lifetime);
    });
```

pranavkm: Makes it easier if the order ever changes.

test/Microsoft.Extensions.Localization.Tests/LocalizationServiceCollectionExtensionsTest.cs
```diff
@@ -21,15 +21,23 @@ public void AddLocalization_AddsNeededServices()
...
 
-            Assert.Equal(typeof(IStringLocalizerFactory), services[0].ServiceType);
-            Assert.Equal(typeof(ResourceManagerStringLocalizerFactory), services[0].ImplementationType);
+            Assert.Equal(typeof(IOptions<>), services[0].ServiceType);
```
kichalla: Ideally I think we do not want to test the services added by options itself as it makes test easy to break if Options decides to change the services (ex: adds/deletes a service etc.). I think we should just check for the services that localization adds itself. (ex: https://github.com/aspnet/Localization/pull/221)

JunTaoLuo: Is Options a required service for Localization though? If it is required to be there, we should make sure it's added when `AddLocalization` is called no? @javiercn?

JunTaoLuo:
#221 will be the approach we take. Closing

PR #221, opened 2016-03-01
Fix test failures

kichalla:
@Eilon @javiercn @muratg


muratg:
:shipit:

Eilon:
:shipit:

JunTaoLuo:
@kichalla which approach should we go with, this one or #220?

javiercn:
:shipit:

Eilon:
@JunTaoLuo I like the approach here a bit more. But could have gone either way.

PR #231, opened 2016-03-27
Fix package metadata

Eilon:
@ryanbrandenburg @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNetCore.Localization/project.json
+++ b/src/Microsoft.AspNetCore.Localization/project.json
@@ -1,6 +1,10 @@
 {
   "version": "1.0.0-*",
-  "description": "Middleware for automatically applying culture information to HTTP requests.",
+  "description": "ASP.NET Core middleware for automatically applying culture information to HTTP requests. Culture information can be specified in the HTTP header, query string, cookie, or custom source.",
+  "tags": [
+    "aspnetcore",
+    "localization"
+  ],
   "repository": {
     "type": "git",
     "url": "https://github.com/aspnet/localization"
@@ -20,7 +24,7 @@
     "Microsoft.Extensions.Options": "1.0.0-*"
   },
   "frameworks": {
-    "net451": {},
+    "net451": { },
     "netstandard1.3": {
       "imports": [
         "dotnet5.4"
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Localization/project.json
```diff
@@ -1,6 +1,10 @@
 {
   "version": "1.0.0-*",
-  "description": "Middleware for automatically applying culture information to HTTP requests.",
+  "description": "ASP.NET Core middleware for automatically applying culture information to HTTP requests. Culture information can be specified in the HTTP header, query string, cookie, or custom source.",
```
ryanbrandenburg: Do we really want to get so specific about how culture information can be specified here? Seems like a deeper dive that might more appropriate for docs to me, and like a thing we would forget to update if we added or removed ways to specify localization.

src/Microsoft.Extensions.Localization.Abstractions/project.json
```diff
@@ -1,6 +1,9 @@
 {
   "version": "1.0.0-*",
-  "description": "Abstractions of application localization services.",
+  "description": "Abstractions of application localization services.\r\nCommonly used types:\r\nMicrosoft.Extensions.Localization.IStringLocalizer\r\nMicrosoft.Extensions.Localization.IStringLocalizer<T>",
```
ryanbrandenburg: Is "Commonly used types" a convention we follow for all Abstractions packages?

Eilon: It's something I've been doing when there are some especially interesting types in a package. CoreFX team has been doing this and it seems useful to me. It's not specific to abstractions packages.

src/Microsoft.AspNetCore.Localization/project.json
Eilon: I certainly hope we never remove any of these ways - that would be a breaking change. And if we add them, the comment remains correct. The goal is for the description to be sufficient to understand what's in the package and why you'd care to install it. Or, similarly, to understand whether it's ok to remove a reference to the package.

src/Microsoft.Extensions.Globalization.CultureInfoCache/project.json
```diff
@@ -1,6 +1,10 @@
 {
   "version": "1.0.0-*",
   "description": "Provides cached instances of CultureInfo using a generated list of known culture names for use in scenarios where unbounded CultureInfo creation is undesirable.",
+  "tags": [
+    "globalization",
```
muratg: Localization project.json also has "aspnetcore" in the tags. Consider adding it here too.

Eilon: This one isn't for ASP.NET (no AspNet in the name).

PR #235, opened 2016-03-31
Resolve test application paths more robustly

JunTaoLuo:
This will fix the full framework tests launched on Mono. The relative paths were broken since we disable app domains and need to resolve paths from xunit.runner.exe. Resolving from the repo root folder is more robust.

cc @pranavkm @BrennanConroy


JunTaoLuo:
This should resolve https://github.com/aspnet/Coherence-Signed/issues/224

BrennanConroy:
:shipit: get rid of those ......'s

PR #241, opened 2016-05-02
Dispose HttpClient

BrennanConroy:


muratg:
What was the issue (i.e. what does this fix?)

BrennanConroy:
Nothing really, we just aren't disposing the HttpClient in a lot of our tests. It might prevent issues on Mac with too many watched files. Which is happening in Entropy right now

Eilon:
:shipit: , please merge when you have a chance. Thanks!

PR #261, opened 2016-07-01
Switch async void to async Task

pranavkm:
cc @rynowak


davidfowl:
yass

rynowak:
Might be worth doing a global search in Universe. I fixed a few of these in MVC a little while back.

rynowak:
:shipit:

pranavkm:
Yup, I think there's a few more in this repo. Additionally a bunch more in stale repos that I've been pestering @Eilon to delete.

PR #262, opened 2016-07-01
ResourceManagerStringLocalizer.GetAllString without default Resource

### LINKED ISSUE #257: _localizer.GetAllStrings() throw MissingManifestResourceException: No manifests exist for the current culture.
In my project, I try to get all the localized strings to be used in javascript. But it's throw exception when I  call _localizer.GetAllStrings().ToList(); although the _localizer.GetAllStrings("Some string"); work fine.

Am I missing something?

ryanbrandenburg:
Fixes #257 by making GetAllStrings check the satellite assemblies for non-default language resources.

### DIFF
--- a/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
+++ b/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
@@ -6,6 +6,8 @@ using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Resources;
 using Microsoft.Extensions.Localization.Internal;
@@ -249,15 +251,54 @@ namespace Microsoft.Extensions.Localization
             return resourceStreamName;
         }
 
+        private const string AssemblyElementDelimiter = ", ";
+
+        private string ApplyCultureToAssembly(string assemblyFullName, CultureInfo culture)
+        {
+            var assemblyNameElements = _resourceAssemblyWrapper.FullName
+                .Split(new string[] { AssemblyElementDelimiter }, StringSplitOptions.RemoveEmptyEntries);
+            IDictionary<string, string> dict = new Dictionary<string, string>();
+
+            foreach (var assemblyNameElement in assemblyNameElements)
+            {
+                var parts = assemblyNameElement.Split(new char[] { '=' });
+
+                if (parts.Count() == 1)
+                {
+                    parts[0] = parts[0] + ".resources";
+                }
+
+                dict.Add(parts[0], parts.Count() == 2 ? parts[1] : null);
+            }
+
+            // Overwrite any existing culture
+            dict["Culture"] = culture.Name == string.Empty ? "neutral" : culture.Name;
+
+            var elements = dict.Select(kvp => kvp.Value == null ? kvp.Key : $"{kvp.Key}={kvp.Value}");
+
+            return string.Join(AssemblyElementDelimiter, elements);
+        }
+
         private IList<string> GetResourceNamesForCulture(CultureInfo culture)
         {
-            var resourceStreamName = GetResourceStreamName(culture);
+            var assemblyName = ApplyCultureToAssembly(_resourceAssemblyWrapper.FullName, culture);
 
+            Assembly assembly;
+            try
+            {
+                assembly = Assembly.Load(new AssemblyName(assemblyName));
+            }
+            catch (FileNotFoundException)
+            {
+                return null;
+            }
+
+            var resourceStreamName = GetResourceStreamName(culture);
             var cacheKey = $"assembly={_resourceAssemblyWrapper.FullName};resourceStreamName={resourceStreamName}";
 
-            var cultureResourceNames = _resourceNamesCache.GetOrAdd(cacheKey, _ =>
+            return _resourceNamesCache.GetOrAdd(cacheKey, _ =>
             {
-                using (var cultureResourceStream = _resourceAssemblyWrapper.GetManifestResourceStream(resourceStreamName))
+                using (var cultureResourceStream = assembly.GetManifestResourceStream(resourceStreamName))
                 {
                     if (cultureResourceStream == null)
                     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -249,15 +251,54 @@ private string GetResourceStreamName(CultureInfo culture)
...
+
+            Assembly assembly;
+            try
+            {
+                assembly = Assembly.Load(new AssemblyName(assemblyName));
+            }
+            catch (FileNotFoundException)
+            {
+                return null;
+            }
```
dougbu: Failures here should also be cached. Move the whole `assembly` part into the lambda below. That's closer to its first use as well.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -249,15 +251,54 @@ private string GetResourceStreamName(CultureInfo culture)
             return resourceStreamName;
         }
 
+        private const string AssemblyElementDelimiter = ", ";
+
+        private string ApplyCultureToAssembly(string assemblyFullName, CultureInfo culture)
+        {
+            var assemblyNameElements = assemblyFullName
+                .Split(new string[] { AssemblyElementDelimiter }, StringSplitOptions.RemoveEmptyEntries);
```
dougbu: Why create a new array every time `ApplyCultureToAssembly()` is called?

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -249,15 +251,54 @@ private string GetResourceStreamName(CultureInfo culture)
             return resourceStreamName;
         }
 
+        private const string AssemblyElementDelimiter = ", ";
```
dougbu: Move to the top of the file.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -249,15 +251,54 @@ private string GetResourceStreamName(CultureInfo culture)
...
+
+        private string ApplyCultureToAssembly(string assemblyFullName, CultureInfo culture)
+        {
+            var assemblyNameElements = assemblyFullName
+                .Split(new string[] { AssemblyElementDelimiter }, StringSplitOptions.RemoveEmptyEntries);
+            IDictionary<string, string> dict = new Dictionary<string, string>();
```
dougbu: `var`

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -249,15 +251,54 @@ private string GetResourceStreamName(CultureInfo culture)
             return resourceStreamName;
         }
 
+        private const string AssemblyElementDelimiter = ", ";
+
+        private string ApplyCultureToAssembly(string assemblyFullName, CultureInfo culture)
```
dougbu: Looks like a very allocation-way to insert `.resources` after the assembly name and optionally to add or update the culture. Result will be cached once you've applied my suggestions below. But still, ...

Also needs comments.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -249,15 +251,54 @@ private string GetResourceStreamName(CultureInfo culture)
...
+        {
+            var assemblyNameElements = assemblyFullName
+                .Split(new string[] { AssemblyElementDelimiter }, StringSplitOptions.RemoveEmptyEntries);
+            IDictionary<string, string> dict = new Dictionary<string, string>();
+
+            foreach (var assemblyNameElement in assemblyNameElements)
+            {
+                var parts = assemblyNameElement.Split(new char[] { '=' });
+
+                if (parts.Count() == 1)
```
dougbu: `parts.Length` and then remove the `using` for Linq.

dougbu:
:watch:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
```diff
@@ -249,15 +251,54 @@ private string GetResourceStreamName(CultureInfo culture)
             return resourceStreamName;
         }
 
+        private const string AssemblyElementDelimiter = ", ";
+
+        private string ApplyCultureToAssembly(string assemblyFullName, CultureInfo culture)
+        {
+            var assemblyNameElements = assemblyFullName
```
hishamco: You can use this instead
```csharp
private static readonly char[] _assemblyElementSeparator = new[] { ',' , ' ' };
```
after that you can use the array in ```Split``` without creating an array everytime

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
hishamco: You can use this instead
```csharp
private static readonly char[] _assemblyElementSeparator = new[] { ',' , ' ' };
```
after that you can use the array in ```Split``` without creating an array everytime

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: You are correct, in addition I should be using `assemblyName` in the cache key rather than `_resourceAssemblyWrapper.FullName` so that we cache the correct thing.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs
ryanbrandenburg: This is a workaround for dotnet/coreclr#6123 so I'm hesitant to go too crazy with optimizing here but I suppose I could do `string.IndexOf`'s and `StringBuilder.Insert`'s instead unless you had something else in mind.

dougbu: > `string.IndexOf`'s and `StringBuilder.Insert`'s instead

That'll work.

ryanbrandenburg:
Closing this in favor of the experiment.

PR #265, opened 2016-07-11
Fix Class libraries on ResourceManagerStringLocalizerFactory

ryanbrandenburg:
This is a hypothetical fix to #157. @DamianEdwards will probably want to check at least the changes in `ResourceManagerStringLocalizerFactory` to ensure they don't break any scenarios that I didn't think of/know about.


global.json
```diff
@@ -1,3 +1,3 @@
 {
-  "projects": [ "src" ]
+  "projects": [ "src", "samples" ]
```
javiercn: We should have a functional test project instead of reusing the samples.

samples/MyLibrary/Class1.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Localization;
+
+namespace MyLibrary
+{
+    public class Class1
+    {
+        private IStringLocalizer<Class1> _localizer;
+
+        public Class1()
```
javiercn: Do not use default names like class1

samples/MyLibrary/Class1.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var services = new ServiceCollection();
```
javiercn: You shouldn't do something artificial like this, specially if it's on a sample. Follow the patterns that we would recommend a customer to follow.

samples/MyLibrary/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
```
javiercn: No need for this file

samples/MyLibrary/OutsiteNamespaceClass.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Internal;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Localization;
+
+namespace OtherNamespace
+{
+    public class OutsideNamespaceClass
```
DamianEdwards: Class name doesn't match file name.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -66,11 +66,23 @@ public IStringLocalizer Create(Type resourceSource)
...
-            var baseName = string.IsNullOrEmpty(_resourcesRelativePath)
-                ? typeInfo.FullName
-                : _hostingEnvironment.ApplicationName + "." + _resourcesRelativePath
-                    + TrimPrefix(typeInfo.FullName, _hostingEnvironment.ApplicationName + ".");
+            string baseName;
+
+            if (_hostingEnvironment.ApplicationName == null)
+            {
+                var assemblyName = assembly.GetName().Name + ".";
+                //We're working in a library
```
DamianEdwards: Leading space

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -66,11 +66,23 @@ public IStringLocalizer Create(Type resourceSource)
...
+                baseName = assemblyName + _resourcesRelativePath + TrimPrefix(typeInfo.FullName, assemblyName);
+            }
+            else
+            {
+                //Inside the main assembly
```
DamianEdwards: Leading space

samples/MyLibrary/Class1.cs
DamianEdwards: Would definitely like to see this using `IStringLocalizer<Class1>` via ctor injection to ensure the recommended end-to-end actually works.

DamianEdwards:
:watch:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -66,11 +66,23 @@ public IStringLocalizer Create(Type resourceSource)
...
-            // Re-root the base name if a resources path is set
-            var baseName = string.IsNullOrEmpty(_resourcesRelativePath)
-                ? typeInfo.FullName
-                : _hostingEnvironment.ApplicationName + "." + _resourcesRelativePath
-                    + TrimPrefix(typeInfo.FullName, _hostingEnvironment.ApplicationName + ".");
+            string baseName;
+
+            if (_hostingEnvironment.ApplicationName == null)
+            {
+                var assemblyName = assembly.GetName().Name + ".";
```
javiercn: I would name that assemblyPathPrefix or something like that. When you say assemblyName I think of A.B.C or SomeAssembly, not something that ends with a dot and that makes my brain tweak when I see you concatenating paths below. (I had to go back to this part of the code, that's why I'm pointing it out).

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -66,11 +66,23 @@ public IStringLocalizer Create(Type resourceSource)
...
+                //We're working in a library
+                baseName = assemblyName + _resourcesRelativePath + TrimPrefix(typeInfo.FullName, assemblyName);
```
javiercn: Don't know why I can't comment on the appropiate line, but in https://github.com/aspnet/Localization/pull/265/files#diff-ec3f46a4394df7156f745d976f053127R49 can you check if you can have a resourcesRelativePath ending in . in the options? In that case it could end up with two . on it. (It's possibly a separate issue)

samples/MyLibrary/Class1.cs
ryanbrandenburg: I'm not sure that's possible to do in general due to ResourcePath. If the ResourcePath being used by your library isn't the same as the one being used by your main app how are you going to know where to look for it? That information is normally kept in the `ResourceManagerStringLocalizerFactory`, but we can't rely on it being the same between assemblies. That may mean that this whole exercise is in vain.

ryanbrandenburg:
Per the in-person conversation with @javiercn and @DamianEdwards I'm closing this PR (but leaving the branch up) until we can have a design meeting to discuss all the various use-cases we need to support and how they might best be achieved without a breaking change, because we've all agreed that solving this is a bit more involved than this PR.

Issue #214, opened 2016-02-18
Question: How can you use a different resource file in a satellite assembly from a class library project?


ghost:
I my case, the default Resources.resx file from the class library is always used. The localized versions like Resource.de.resx are ignored.

Any hint?

hishamco:
I think it's duplicate for #154

pepevc:
My class library is a "dnx" assembly, not a regular one.

hishamco:
@Eilon is this blocked by #157

Eilon:
@pepevc what exactly is the scenario here? Can you provide more info about what is in each project, what the code is trying to do, what doesn't work, and what you were expecting?

pepevc:
Pick for instance https://github.com/aspnet/identity. That project contains
a resx file for error messages. I would like the localization framework to
pick the correct culture resx file automatically.

Eilon:
@pepevc so you have a solution like this:

1. **MyLibrary**
 - Has Class1.cs with a method `GetData()` that loads resources from:
 - Class1.resx (has neutral en-US resources)
 - Class1.fr-fr.resx (has French resources)
 - Class1.ja-jp.resx (has Japanese resources)
2. **MyApp**
 - Has Program.cs that calls `Class1.GetData()`
 
And you're saying that when it called `GetData` that it always picks up English, no matter what?

pepevc:
@Eilon yes, basically that's it. I am following the Identity project as an example. Have a look at their IdentityErrorDescriber class and the AssemblyInfo.cs.

I reproduced something similar with with a dnx class library project but my aspnet project fails to make my class library load anything besides the main resx file.

Not a big issue as I will end up doing the localization in the webapp, but the bug? is there.

Eilon:
@pepevc thanks for the extra info!

@ryanbrandenburg can you take a look?

ryanbrandenburg:
@DamianEdwards could you opine on this? It was my impression that not having resources from satellite assemblies was by design.

MrWinky:
"_It was my impression that not having resources from satellite assemblies was by design_."

I hope not!  Because is a major problem that breaks  SOLID architecture principles (for loose coupling and tight cohesion).  

For example:

1.   EF 7 is kept in it's own project.  MyProject.Data.EntityFramework
2.  Business logic layer objects which use the strings  MyProject.BLL.ViewModels
2.  The web portal (with startup.cs) is in it's own project MyProject.Portal

This is pretty standard N tier architecture.   

[...]

PR #266, opened 2016-07-27
An example ResourceManagerStringLocalizerFactory which uses libraries

ryanbrandenburg:
If everyone's ok with the general shape of things here I'll move `ClassLibraryResourceManagerStringLocalizerFactory.cs` into the Entropy repo and give it a sample.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ nuget.exe
 *.ipch
 *.sln.ide
 *launchSettings.json
+**/Resources/*.Designer.cs
--- a/samples/LocalizationSample/Startup.cs
+++ b/samples/LocalizationSample/Startup.cs
@@ -37,7 +37,8 @@ namespace LocalizationSample
 #if !NETCOREAPP1_0
             supportedCultures.Add(new CultureInfo("zh-CHT"));
 #endif
-            var options = new RequestLocalizationOptions {
+            var options = new RequestLocalizationOptions
+            {
                 DefaultRequestCulture = new RequestCulture("en-US"),
                 SupportedCultures = supportedCultures,
                 SupportedUICultures = supportedCultures
--- a/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
+++ b/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
@@ -51,6 +51,70 @@ namespace Microsoft.Extensions.Localization
         }
 
         /// <summary>
+        /// Gets the resource prefix used to look up the resource.
+        /// </summary>
+        /// <param name="typeInfo">The type of the resource to be looked up.</param>
+        /// <returns>The prefix for resource lookup.</returns>
+        protected virtual string GetResourcePrefix(TypeInfo typeInfo)
+        {
+            if (typeInfo == null)
+            {
+                throw new ArgumentNullException(nameof(typeInfo));
+            }
+
+            return GetResourcePrefix(typeInfo, _hostingEnvironment.ApplicationName, _resourcesRelativePath);
+        }
+
+        /// <summary>
+        /// Gets the resource prefix used to look up the resource.
+        /// </summary>
+        /// <param name="typeInfo">The type of the resource to be looked up.</param>
+        /// <param name="baseNamespace">The base namespace of the application.</param>
+        /// <param name="resourcesRelativePath">The folder containing all resources.</param>
+        /// <returns>The prefix for resource lookup.</returns>
+        /// <remarks>
+        /// For the type "Sample.Controllers.Home" if there's a resourceRelativePath return
+        /// "Sample.Resourcepath.Controllers.Home" if there isn't one then it would return "Sample.Controllers.Home".
+        /// </remarks>
+        protected virtual string GetResourcePrefix(TypeInfo typeInfo, string baseNamespace, string resourcesRelativePath)
+        {
+            if (typeInfo == null)
+            {
+                throw new ArgumentNullException(nameof(typeInfo));
+            }
+
+            if (string.IsNullOrEmpty(baseNamespace))
+            {
+                throw new ArgumentNullException(nameof(baseNamespace));
+            }
+
+            return string.IsNullOrEmpty(resourcesRelativePath)
+                ? typeInfo.FullName
+                : baseNamespace + "." + resourcesRelativePath + TrimPrefix(typeInfo.FullName, baseNamespace + ".");
+        }
+
+        /// <summary>
+        /// Gets the resource prefix used to look up the resource.
+        /// </summary>
+        /// <param name="baseResourceName">The name of the resource to be looked up</param>
+        /// <param name="baseNamespace">The base namespace of the application.</param>
+        /// <returns>The prefix for resource lookup.</returns>
+        protected virtual string GetResourcePrefix(string baseResourceName, string baseNamespace)
+        {
+            if (string.IsNullOrEmpty(baseResourceName))
+            {
+                throw new ArgumentNullException(nameof(baseResourceName));
+            }
+
+            var locationPath = baseNamespace == _hostingEnvironment.ApplicationName ?
+                baseNamespace + "." + _resourcesRelativePath :
+                baseNamespace + ".";
+            baseResourceName = locationPath + TrimPrefix(baseResourceName, baseNamespace + ".");
+
+            return baseResourceName;
+        }
+
+        /// <summary>
         /// Creates a <see cref="ResourceManagerStringLocalizer"/> using the <see cref="Assembly"/> and
         /// <see cref="Type.FullName"/> of the specified <see cref="Type"/>.
         /// </summary>
@@ -67,10 +131,7 @@ namespace Microsoft.Extensions.Localization
             var assembly = typeInfo.Assembly;
 
             // Re-root the base name if a resources path is set
-            var baseName = string.IsNullOrEmpty(_resourcesRelativePath)
-                ? typeInfo.FullName
-                : _hostingEnvironment.ApplicationName + "." + _resourcesRelativePath
-                    + TrimPrefix(typeInfo.FullName, _hostingEnvironment.ApplicationName + ".");
+            var baseName = GetResourcePrefix(typeInfo);
 
             return _localizerCache.GetOrAdd(baseName, _ =>
                 new ResourceManagerStringLocalizer(
@@ -96,7 +157,7 @@ namespace Microsoft.Extensions.Localization
 
             location = location ?? _hostingEnvironment.ApplicationName;
 
-            baseName = location + "." + _resourcesRelativePath + TrimPrefix(baseName, location + ".");
+            baseName = GetResourcePrefix(baseName, location);
 
             return _localizerCache.GetOrAdd($"B={baseName},L={location}", _ =>
             {
--- a/Localization.sln
+++ b/Localization.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{FB313677-BAB3-4E49-8CDB-4FA4A9564767}"
 EndProject


hishamco:
IMHO the Entropy repo is the suitable place for such thing, nothing but why you change some of the source files while it's an example?!!

ryanbrandenburg:
@hishamco this is a reaction to #214, we're trying to make it easier to sub-class `ResourceManagerStringLocalizerFactory` so that people can write bespoke solutions to their Resource needs.

CC @DamianEdwards, @Eilon, and @javiercn for review and to see if we could further limit changes to `ResourceManagerStringLocalizerFactory` and still make sub-classing easy.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -50,6 +50,29 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
             }
         }
 
+        protected virtual string GetKey(TypeInfo typeInfo)
```
DamianEdwards: Can we check if "key" is the appropriate noun here? This is about returning the name of the embedded resource, yes? Let's check what the framework APIs refer to this as.

ryanbrandenburg: `ResourceManager` calls it "baseName".  I think that's a little obscure in our use case, maybe "GetResourceBaseName"? 100% agree now that you've pointed it out that "Key" is a terrible noun here.

ryanbrandenburg:
@pranavkm and @kichalla could I get you guys to give this a once over since @DamianEdwards is on vacation?

.gitignore
```diff
@@ -30,4 +30,5 @@ nuget.exe
 *.*sdf
 *.ipch
 *.sln.ide
-*launchSettings.json
\ No newline at end of file
+*launchSettings.json
+*.Designer.cs
```
pranavkm: This would be problematic for regular designer files (which the project has). Could you specify a path for this?

samples/LocalizationSample/ClassLibraryResourceManagerStringLocalizerFactory.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Collections.Generic;
```
pranavkm: License

samples/LocalizationSample/ClassLibraryResourceManagerStringLocalizerFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// and will properly handle the resources of ClassLibraries.
+    /// </summary>
+    public class ClassLibraryStringLocalizerFactory : ResourceManagerStringLocalizerFactory
+    {
+        private IReadOnlyDictionary<string, string> _resourcePathMappings;
+
+        public ClassLibraryStringLocalizerFactory(
+            IHostingEnvironment hostingEnvironment,
+            IOptions<LocalizationOptions> localizationOptions,
+            IReadOnlyDictionary<string, string> resourcePathMappings)
```
pranavkm: `IDictionary<,>`.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
pranavkm: `GetKeyPrefix` \ `GetBaseName`?

samples/LocalizationSample/Startup.cs
```diff
@@ -10,18 +10,38 @@
...
 
 namespace LocalizationSample
 {
     public class Startup
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddLocalization(options => options.ResourcesPath = "My/Resources");
+            services.AddOptions();
+            services.TryAddSingleton(typeof(IStringLocalizerFactory), (serviceProvider) =>
```
pranavkm: Could we configure `ClassLibraryStringLocalizerFactory` via options? Seems gross to do `GetRequiredService`

samples/LocalizationSample/Startup.cs
```diff
@@ -10,18 +10,38 @@
...
+                    };
+                    return new ClassLibraryStringLocalizerFactory(hostingEnvironment, localizationOptions, resourcePaths);
+                });
+            services.TryAddTransient(typeof(IStringLocalizer<>), typeof(StringLocalizer<>));
         }
 
-        public void Configure(IApplicationBuilder app, IStringLocalizer<Startup> SR)
+        public void Configure(
+            IApplicationBuilder app,
+            IStringLocalizer<Startup> SR,
```
pranavkm: `startupStringLocalizer`.

samples/MyLibrary/OutsideNamespaceClass.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.Extensions.Localization;
+
+namespace OtherNamespace
+{
+    public class OutsideNamespaceClass
```
pranavkm: `OtherNamespaceClass`. `OutsideNamespace` sounds like it's a type in the global namespace

pranavkm:
:watch:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -50,6 +50,29 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
+        {
+            return GetKey(typeInfo, _hostingEnvironment.ApplicationName, _resourcesRelativePath);
+        }
+
+        protected virtual string GetKey(TypeInfo typeInfo, string location, string resourcesRelativePath)
+        {
+            return string.IsNullOrEmpty(resourcesRelativePath)
```
kichalla: Could you add a comment on how these keys become in the end after the transformation(also in the other overload)? not immediately clear what this is doing unless i spend some time calculating the parts.

samples/LocalizationSample/Startup.cs
kichalla: I agree with @pranavkm

samples/LocalizationSample/Startup.cs
```diff
@@ -10,18 +10,38 @@
...
+using MyLibrary;
+using OtherNamespace;
 
 namespace LocalizationSample
 {
     public class Startup
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddLocalization(options => options.ResourcesPath = "My/Resources");
```
kichalla: you need not remove this actually(but yeah remove the `options` thingy here)...if you register the `ClassLibraryStringLocalizerFactory` either before or after this call, it would replace the default `ResourceManagerStringLocalizerFactory`.  This is because we always do a `TryAdd` to avoid overriding user provided implementation: https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/LocalizationServiceCollectionExtensions.cs#L64

samples/LocalizationSample/Startup.cs
```diff
@@ -10,18 +10,38 @@
 using Microsoft.AspNetCore.Localization;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Options;
+using MyLibrary;
+using OtherNamespace;
```
javiercn: Find appropiate names

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
kichalla: These need docs as users can override this functionality (instead of the comment i was asking for you can probably add a remark to the doc comment itself)

kichalla:
:watch:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: Talked in person, settled on `GetResourcePrefix`.

samples/LocalizationSample/ClassLibraryResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: I guess this one's moot if I'm moving it into an options class.

pranavkm: fc 550x550 grass_green u2

ryanbrandenburg:
:up::date:

samples/LocalizationSample/Startup.cs
```diff
@@ -12,35 +13,35 @@
...
-                    var localizationOptions = serviceProvider.GetRequiredService<IOptions<LocalizationOptions>>();
-                    var resourcePaths = new Dictionary<string, string>
-                    {
-                        {"MyLibrary", "Resources"},
-                        {"LocalizationSample", "My/Resources" }
-                    };
-                    return new ClassLibraryStringLocalizerFactory(hostingEnvironment, localizationOptions, resourcePaths);
-                });
-            services.TryAddTransient(typeof(IStringLocalizer<>), typeof(StringLocalizer<>));
+                    {"HelloServices", "Resources"},
```
Eilon: Formatting: should have spaces before and after each `{` and `}`.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -50,20 +50,29 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
         }
 
-        protected virtual string GetKey(TypeInfo typeInfo, string location, string resourcesRelativePath)
+        /// <summary>
+        /// Constructs the baseName of a resource for use in retrieving it.
+        /// </summary>
+        /// <param name="typeInfo"></param>
+        /// <param name="location"></param>
+        /// <param name="resourcesRelativePath"></param>
+        /// <remarks>If we have a resourcePath we should return "AppName.Resourcepath.(TypeName.trim())"
```
Eilon: I know this is just a sample, but docs should never say `we`. What is it that should have a resourcePath?

kichalla: This piece is actually product code…could we show an example (like with actual paths)? for example, something like here: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor/RazorViewEngineOptions.cs#L57-L59

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -50,6 +50,38 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
             }
         }
 
+        protected virtual string GetResourcePrefix(TypeInfo typeInfo)
+        {
+            return GetResourcePrefix(typeInfo, _hostingEnvironment.ApplicationName, _resourcesRelativePath);
+        }
+
+        /// <summary>
+        /// Constructs the baseName of a resource for use in retrieving it.
```
pranavkm: Construct the resource prefix used the lookup a resource.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -50,6 +50,38 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
+        /// </summary>
+        /// <param name="typeInfo"></param>
```
pranavkm: Needs docs

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -50,6 +50,38 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
             }
         }
 
+        protected virtual string GetResourcePrefix(TypeInfo typeInfo)
```
pranavkm: `protected` methods require docs.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -50,6 +50,38 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
+        {
+            return GetResourcePrefix(typeInfo, _hostingEnvironment.ApplicationName, _resourcesRelativePath);
```
pranavkm: null check `typeInfo`

pranavkm: Similarly for other methods.

pranavkm:
:watch:  for docs. Code changes look good.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
Eilon: Oh duh of course it's product code 😄 I was jumping too much between two files...

ryanbrandenburg:
:up::date:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -51,6 +51,71 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
         }
 
         /// <summary>
+        /// Construct the resource prefix used to look up the resource.
+        /// </summary>
+        /// <param name="typeInfo">The type of the Resource to be looked up.</param>
```
pranavkm: > The `<see cref="TypeInfo" />` of the resource to be looked up.

Resource doesn't need to be capitalized.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -51,6 +51,71 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
+        /// <returns>A <see cref="string"/> representing the Resource name as it is stored without culture.</returns>
```
pranavkm: > The prefix for the lookup.

or something like that

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -51,6 +51,71 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
         }
 
         /// <summary>
+        /// Construct the resource prefix used to look up the resource.
```
pranavkm: ~~Construct~~ Gets

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -51,6 +51,71 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
+                throw new ArgumentNullException(nameof(typeInfo));
+            }
+
+            return GetResourcePrefix(typeInfo, _hostingEnvironment.ApplicationName, _resourcesRelativePath);
+        }
+
+        /// <summary>
+        /// Construct the resource prefix used to look up the resource.
+        /// </summary>
+        /// <param name="typeInfo">The type of the Resource to be looked up.</param>
```
pranavkm: lower cased resource.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -51,6 +51,71 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
+        /// <param name="baseNamespace">The base namespace of the application.</param>
+        /// <param name="resourcesRelativePath">The folder containing all resources.</param>
+        /// <remarks>
+        /// For the type "Sample.Controllers.Home" if there's a resourceRelativePath return
+        /// "Sample.Resourcepath.Controllers.Home" if there isn't one then it would return "Sample.Controllers.Home".
+        /// </remarks>
+        /// <returns>A <see cref="string"/> representing the Resource name as it is stored without culture.</returns>
```
pranavkm: More of a preference, `remarks` usually go at the end of the doc summary. Also, same issue as the previous returns and applies to the next method

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -51,6 +51,71 @@ public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
...
+
+            if (string.IsNullOrEmpty(baseNamespace))
+            {
+                throw new ArgumentNullException(nameof(baseNamespace));
+            }
+
+            return string.IsNullOrEmpty(resourcesRelativePath)
+                ? typeInfo.FullName
+                : baseNamespace + "." + resourcesRelativePath
+                    + TrimPrefix(typeInfo.FullName, baseNamespace + ".");
```
pranavkm: Could this go on the previous line?

pranavkm:
:shipit: once comments are addressed.

PR #274, opened 2016-08-09
Remove package listings

pranavkm:


pranavkm:
cc @Eilon - we don't consistently list packages in most repos. I'd prefer removing this (or limiting it to listing the version in nuget.org). Thoughts?

Eilon:
@DamianEdwards wanted this. Why remove it?

pranavkm:
Well, we should add it to all the repos in that case (and update the feeds when we migrate).. Right now, there are 4 repos under aspnet which have feed details.

Eilon:
Do the shields.io images work with dotnet.myget.org? But anyway, I don't think this needs to be an all-or-none.

pranavkm:
Fair enough. Regardless, we shouldn't point people to untested bits. Could we change this to just the signed \ tested build + nuget.org?

Eilon:
Yeah I agree there are a few to many package variations listed here. Feel free to trim as needed.

pranavkm:
Updated

README.md
```diff
@@ -4,10 +4,10 @@ AppVeyor: [![AppVeyor](https://ci.appveyor.com/api/projects/status/omn0l2l3mfhfj
 
 Travis:   [![Travis](https://travis-ci.org/aspnet/Localization.svg?branch=dev)](https://travis-ci.org/aspnet/Localization)
 
-| Package | aspnetvolatiledev | [aspnetcidev](https://myget.org/gallery/aspnetcidev) | [aspnetvnext](https://myget.org/gallery/aspnetvnext) | [NuGet](https://nuget.org) |
+| Package | [aspnet-core](https://myget.org/gallery/aspnetcore-dev) | [NuGet](https://nuget.org) |
```
Eilon: dotnet.myget.org?

Eilon:
One comment, then :shipit:

pranavkm:
Merged in https://github.com/aspnet/Localization/commit/09123cb349c03c497cff10e5105f87e18da64968

PR #280, opened 2016-09-04
Increase .travis.yml consistency between repos

dougbu:
- aspnet/Universe#349
- minimize `dotnet` setup time; no need for caching
- no need for libuv v1.4.2 w/ current Kestrel packages


dougbu:
f650de8

Issue #281, opened 2016-09-12
Application parts, MissingManifestResourceException: The manifest 'Xxx.Yyy.Zzz.en.resources' was not found


DmitrySikorsky:
I'm working on modular application and trying to add application part (assembly 'Module') to the main application (assembly 'Application'). Also I have direct link to that project in Application's project.json. 

Module contains localized views (using IViewLocalizer) and resx files in the Resources folder. Values from the resx not shown. If I try to call IViewLocalizer's GetAllStrings method in the View I will get next exception:

MissingManifestResourceException: The manifest 'Application.Resources.Views.Cultures.Index.en.resources' was not found.

Why does it look for Application.Resources but not for Module.Resource? I have checked and there is Module.Resources.Views.Cultures.Index.en.resources available inside the Module assembly.

Could you please help me with that? Thanks!

hishamco:
I think it's a duplicate for issue #257

DmitrySikorsky:
Maybe. GetManifestResourceNames() returns nothing (except views) for me too. But I know that resource is added because if I add resx like A.B.C.resx in addition to A/B/C.resx I get compilation error that resource id is already exists.

Problem is that I'm not sure that I'm doing this in the right way. Localization works if I have localized views and resx in the Application project, but it doesn't work in the views from separate assembly. (Just does nothing or throughs exception if I call GetAllStrings). Is it enough just to add localized views and resx in the Module project and then use AddApplicationPart to add that assembly?

hishamco:
> Localization works if I have localized views and resx in the Application project, but it doesn't work in the views from separate assembly.

This is seems a bug in 1.1.0-alpha1-21873, please have a look to this issue #275

DmitrySikorsky:
Problem is that I'm using 1.0.0. I will try to create simple demo for you.

Eilon:
We think we have a way of making this fairly easily doable in an app:

1. We'll define a new assembly-level attribute `[assembly: LocalizerResourcePath(string resourcePath)]`
2. Assemblies such as those containing application parts would use this new attribute to specify their root resource path.
3. The default localization system would check the type of the resource being used, see if its containing assembly has this attribute, and if so, use *that* as the resource root path for lookups. If the attribute isn't there, then the current behavior will be used, which is that the localization Options class has its own resource path.

PinpointTownes:
@Eilon great approach (similar to the one I had in mind).

[...]

Issue #275, opened 2016-08-16
update problem : don't support separate library in new update


pMonfared:
i'm using localization in separate class library and it worked in 1.0.0
but don't work after updated to 1.1.0-alpha1-21873

```
public class AccountController : Controller
    {
        private readonly IStringLocalizer _stringLocalizer;
        private readonly IHtmlLocalizer _htmlLocalizer;

        private readonly UserManager<ApplicationUser> _userManager;
        private readonly SignInManager<ApplicationUser> _signInManager;
        private readonly IEmailSender _emailSender;
        private readonly ISmsSender _smsSender;
        private readonly ILogger _logger;

        public AccountController(
            IStringLocalizerFactory stringLocalizerFactory,
            IHtmlLocalizerFactory htmlLocalizerFactory,
            UserManager<ApplicationUser> userManager,
            SignInManager<ApplicationUser> signInManager,
            IEmailSender emailSender,
            ISmsSender smsSender,
            ILoggerFactory loggerFactory)
        {
            _stringLocalizer = stringLocalizerFactory.Create(
        baseName: "Controllers.AccountController" ,
        location: "Bookkhan.ExternalResources" );
            _htmlLocalizer = htmlLocalizerFactory.Create(
                baseName: "Controllers.AccountController" ,
                location: "Bookkhan.ExternalResources" );

            _userManager = userManager;
            _signInManager = signInManager;
            _emailSender = emailSender;
            _smsSender = smsSender;
            _logger = loggerFactory.CreateLogger<AccountController>();
        }
...
```

![untitledsss](https://cloud.githubusercontent.com/assets/10448255/17697731/bd31e866-63cb-11e6-8342-a4695d3e61ee.jpg)

Eilon:
Let's see if there's a regression here and if so, we should make sure we add tests before we fix.

hishamco:
@pMonfared do you have the sample in github repo?

Eilon:
@ryanbrandenburg let's see if we can make this work properly once we do the other fix with the attribute.

ryanbrandenburg:
@pMonfared this should be fixed by #283 which is merged into dev now? Can you give `ResourceLocationAttribute` a try and confirm it fixes your situation?

ryanbrandenburg:
On second thought, what you've said here is pretty much the exact bug we fixed, I'll just close this out and you feel free to reopen it if you're still having trouble.

PR #283, opened 2016-09-16
Class library resources

ryanbrandenburg:
This should allow localization to find resources in a class library by using an Assembly attribute to note the location of the resources folder.

### DIFF
--- a/src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
+++ b/src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
@@ -1,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Provides the location of resources for an Assembly.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+    public class ResourceLocationAttribute : Attribute
+    {
+        /// <summary>
+        /// Creates a new <see cref="ResourceLocationAttribute"/>.
+        /// </summary>
+        /// <param name="resourceLocation">The location of resources for this Assembly.</param>
+        public ResourceLocationAttribute(string resourceLocation)
+        {
+            if (string.IsNullOrEmpty(resourceLocation))
+            {
+                throw new ArgumentNullException(nameof(resourceLocation));
+            }
+
+            ResourceLocation = resourceLocation;
+        }
+
+        /// <summary>
+        /// The location of resources for this Assembly.
+        /// </summary>
+        public string ResourceLocation { get; }
+    }
+}
--- a/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
+++ b/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
@@ -14,12 +14,17 @@ namespace Microsoft.Extensions.Localization
     /// <summary>
     /// An <see cref="IStringLocalizerFactory"/> that creates instances of <see cref="ResourceManagerStringLocalizer"/>.
     /// </summary>
+    /// <remarks>
+    /// <see cref="ResourceManagerStringLocalizerFactory"/> offers multiple ways to set the relative path of
+    /// resources to be used. They are, in order of precedence:
+    /// <see cref="ResourceLocationAttribute"/> -> <see cref="LocalizationOptions.ResourcesPath"/> -> the project root.
+    /// </remarks>
     public class ResourceManagerStringLocalizerFactory : IStringLocalizerFactory
     {
         private readonly IResourceNamesCache _resourceNamesCache = new ResourceNamesCache();
         private readonly ConcurrentDictionary<string, ResourceManagerStringLocalizer> _localizerCache =
             new ConcurrentDictionary<string, ResourceManagerStringLocalizer>();
-        private readonly IHostingEnvironment _hostingEnvironment;
+        private readonly string _applicationName;
         private readonly string _resourcesRelativePath;
 
         /// <summary>
@@ -41,7 +46,7 @@ namespace Microsoft.Extensions.Localization
                 throw new ArgumentNullException(nameof(localizationOptions));
             }
 
-            _hostingEnvironment = hostingEnvironment;
+            _applicationName = hostingEnvironment.ApplicationName;
             _resourcesRelativePath = localizationOptions.Value.ResourcesPath ?? string.Empty;
             if (!string.IsNullOrEmpty(_resourcesRelativePath))
             {
@@ -62,7 +67,7 @@ namespace Microsoft.Extensions.Localization
                 throw new ArgumentNullException(nameof(typeInfo));
             }
 
-            return GetResourcePrefix(typeInfo, _hostingEnvironment.ApplicationName, _resourcesRelativePath);
+            return GetResourcePrefix(typeInfo, new AssemblyName(typeInfo.Assembly.FullName).Name, _resourcesRelativePath);
         }
 
         /// <summary>
@@ -106,9 +111,16 @@ namespace Microsoft.Extensions.Localization
                 throw new ArgumentNullException(nameof(baseResourceName));
             }
 
-            var locationPath = baseNamespace == _hostingEnvironment.ApplicationName ?
-                baseNamespace + "." + _resourcesRelativePath :
-                baseNamespace + ".";
+            if (string.IsNullOrEmpty(baseNamespace))
+            {
+                throw new ArgumentNullException(nameof(baseNamespace));
+            }
+
+            var assemblyName = new AssemblyName(baseNamespace);
+            var assembly = Assembly.Load(assemblyName);
+            var resourceLocation = GetResourcePath(assembly);
+            var locationPath = baseNamespace + "." + resourceLocation;
+
             baseResourceName = locationPath + TrimPrefix(baseResourceName, baseNamespace + ".");
 
             return baseResourceName;
@@ -129,17 +141,12 @@ namespace Microsoft.Extensions.Localization
 
             var typeInfo = resourceSource.GetTypeInfo();
             var assembly = typeInfo.Assembly;
+            var assemblyName = new AssemblyName(assembly.FullName);
+            var resourcePath = GetResourcePath(assembly);
 
-            // Re-root the base name if a resources path is set
-            var baseName = GetResourcePrefix(typeInfo);
+            var baseName = GetResourcePrefix(typeInfo, assemblyName.Name, resourcePath);
 
-            return _localizerCache.GetOrAdd(baseName, _ =>
-                new ResourceManagerStringLocalizer(
-                    new ResourceManager(baseName, assembly),
-                    assembly,
-                    baseName,
-                    _resourceNamesCache)
-            );
+            return _localizerCache.GetOrAdd(baseName, _ => CreateResourceManagerStringLocalizer(assembly, baseName));
         }
 
         /// <summary>
@@ -155,19 +162,69 @@ namespace Microsoft.Extensions.Localization
                 throw new ArgumentNullException(nameof(baseName));
             }
 
-            location = location ?? _hostingEnvironment.ApplicationName;
-
-            baseName = GetResourcePrefix(baseName, location);
+            location = location ?? _applicationName;
 
             return _localizerCache.GetOrAdd($"B={baseName},L={location}", _ =>
             {
-                var assembly = Assembly.Load(new AssemblyName(location));
-                return new ResourceManagerStringLocalizer(
-                    new ResourceManager(baseName, assembly),
-                    assembly,
-                    baseName,
-                    _resourceNamesCache);
+                var assemblyName = new AssemblyName(location);
+                var assembly = Assembly.Load(assemblyName);
+                baseName = GetResourcePrefix(baseName, location);
+
+                return CreateResourceManagerStringLocalizer(assembly, baseName);
             });
+        }
+
+        /// <summary>Creates a <see cref="ResourceManagerStringLocalizer"/> for the given input.</summary>
+        /// <param name="assembly">The assembly to create a <see cref="ResourceManagerStringLocalizer"/> for.</param>
+        /// <param name="baseName">The base name of the resource to search for.</param>
+        /// <returns>A <see cref="ResourceManagerStringLocalizer"/> for the given <paramref name="assembly"/> and <paramref name="baseName"/>.</returns>
+        /// <remarks>This method is virtual for testing purposes only.</remarks>
+        protected virtual ResourceManagerStringLocalizer CreateResourceManagerStringLocalizer(
+            Assembly assembly,
+            string baseName)
+        {
+            return new ResourceManagerStringLocalizer(
+                new ResourceManager(baseName, assembly),
+                assembly,
+                baseName,
+                _resourceNamesCache);
+        }
+
+        /// <summary>
+        /// Gets the resource prefix used to look up the resource.
+        /// </summary>
+        /// <param name="location">The general location of the resource.</param>
+        /// <param name="baseName">The base name of the resource.</param>
+        /// <param name="resourceLocation">The location of the resource within <paramref name="location"/>.</param>
+        /// <returns>The resource prefix used to look up the resource.</returns>
+        protected virtual string GetResourcePrefix(string location, string baseName, string resourceLocation)
+        {
+            // Re-root the base name if a resources path is set
+            return location + "." + resourceLocation + TrimPrefix(baseName, location + ".");
+        }
+
+        /// <summary>Gets a <see cref="ResourceLocationAttribute"/> from the provided <see cref="Assembly"/>.</summary>
+        /// <param name="assembly">The assembly to get a <see cref="ResourceLocationAttribute"/> from.</param>
+        /// <returns>The <see cref="ResourceLocationAttribute"/> associated with the given <see cref="Assembly"/>.</returns>
+        /// <remarks>This method is protected and virtual for testing purposes only.</remarks>
+        protected virtual ResourceLocationAttribute GetResourceLocationAttribute(Assembly assembly)
+        {
+            return assembly.GetCustomAttribute<ResourceLocationAttribute>();
+        }
+
+        private string GetResourcePath(Assembly assembly)
+        {
+            var resourceLocationAttribute = GetResourceLocationAttribute(assembly);
+
+            // If we don't have an attribute assume all assemblies use the same resource location.
+            var resourceLocation = resourceLocationAttribute == null
+                ? _resourcesRelativePath
+                : resourceLocationAttribute.ResourceLocation + ".";
+            resourceLocation = resourceLocation
+                .Replace(Path.DirectorySeparatorChar, '.')
+                .Replace(Path.AltDirectorySeparatorChar, '.');
+
+            return resourceLocation;
         }
 
         private static string TrimPrefix(string name, string prefix)
--- a/src/Microsoft.Extensions.Localization/project.json
+++ b/src/Microsoft.Extensions.Localization/project.json
@@ -22,7 +22,8 @@
     "Microsoft.AspNetCore.Hosting.Abstractions": "1.1.0-*",
     "Microsoft.Extensions.DependencyInjection.Abstractions": "1.1.0-*",
     "Microsoft.Extensions.Localization.Abstractions": "1.1.0-*",
-    "Microsoft.Extensions.Options": "1.1.0-*"
+    "Microsoft.Extensions.Options": "1.1.0-*",
+    "System.Reflection.Extensions": "4.0.1-*"
   },
   "frameworks": {
     "net451": {},
--- a/test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
+++ b/test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System.Collections.Generic;
+using System.Globalization;
+using System.Reflection;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Localization;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Logging;
+
+namespace LocalizationWebsite
+{
+    public class StartupResourcesInClassLibrary
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddLocalization(options => options.ResourcesPath = "Resources");
+        }
+
+        public void Configure(
+            IApplicationBuilder app,
+            ILoggerFactory loggerFactory,
+            IStringLocalizerFactory stringLocalizerFactory)
+        {
+            loggerFactory.AddConsole(minLevel: LogLevel.Warning);
+
+            var supportedCultures = new List<CultureInfo>()
+            {
+                new CultureInfo("en-US"),
+                new CultureInfo("fr-FR")
+            };
+
+            app.UseRequestLocalization(new RequestLocalizationOptions
+            {
+                DefaultRequestCulture = new RequestCulture("en-US"),
+                SupportedCultures = supportedCultures,
+                SupportedUICultures = supportedCultures
+            });
+
+            var noAttributeStringLocalizer = stringLocalizerFactory.Create(typeof(ResourcesClassLibraryNoAttribute.Model));
+            var withAttributeStringLocalizer = stringLocalizerFactory.Create(typeof(ResourcesClassLibraryWithAttribute.Model));
+
+            var noAttributeAssembly = typeof(ResourcesClassLibraryNoAttribute.Model).GetTypeInfo().Assembly;
+            var noAttributeName = new AssemblyName(noAttributeAssembly.FullName).Name;
+            var noAttributeNameStringLocalizer = stringLocalizerFactory.Create(
+                nameof(ResourcesClassLibraryNoAttribute.Model),
+                noAttributeName);
+
+            var withAttributeAssembly = typeof(ResourcesClassLibraryWithAttribute.Model).GetTypeInfo().Assembly;
+            var withAttributeName = new AssemblyName(withAttributeAssembly.FullName).Name;
+            var withAttributeNameStringLocalizer = stringLocalizerFactory.Create(
+                nameof(ResourcesClassLibraryWithAttribute.Model),
+                withAttributeName);
+
+            app.Run(async (context) =>
+            {
+                await context.Response.WriteAsync(noAttributeNameStringLocalizer["Hello"]);
+                await context.Response.WriteAsync(" ");
+                await context.Response.WriteAsync(noAttributeStringLocalizer["Hello"]);
+                await context.Response.WriteAsync(" ");
+                await context.Response.WriteAsync(withAttributeNameStringLocalizer["Hello"]);
+                await context.Response.WriteAsync(" ");
+                await context.Response.WriteAsync(withAttributeStringLocalizer["Hello"]);
+            });
+        }
+    }
+}
--- a/test/LocalizationWebsite/StartupResourcesInFolder.cs
+++ b/test/LocalizationWebsite/StartupResourcesInFolder.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Globalization;
+using System.Reflection;
 using LocalizationWebsite.Models;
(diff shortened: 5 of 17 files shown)


src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -65,12 +66,27 @@ public IStringLocalizer Create(Type resourceSource)
...
+            var resourcePath = _resourcesRelativePath;
+            var baseNamespace = _hostingEnvironment.ApplicationName;
+
+            //This isn't the main assembly
+            if (!string.Equals(assemblyName.Name, _hostingEnvironment.ApplicationName, StringComparison.Ordinal))
+            {
+                var resourceLocationAttribute = assembly.GetCustomAttribute<ResourceLocationAttribute>();
+                // If we don't have an attribute, just assume it's in 'Resources'
+                resourcePath = string.IsNullOrEmpty(resourceLocationAttribute?.ResourceFolder)
+                    ? "Resources"
```
ryanbrandenburg: After thinking about this, I think the right thing to do here, rather than just assuming the folder is "Resources", is to do basically what we do for project resources and leave it null so we get TypeInfo.FullName.

ryanbrandenburg:
@javiercn I believe you're the next most knowledgeable for localization since @dougbu is unavailable.

hishamco:
Copyright & docs are missing in the source files

src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Localization
+{
+    public class ResourceLocationAttribute : Attribute
+    {
+        public ResourceLocationAttribute(string resourceFolder)
+        {
+            ResourceFolder = resourceFolder;
```
hishamco: Why we didn't throw `ArgumentNullException` instead of assuming the resource folder name is "Resources"

src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+        }
+
+        public string ResourceFolder { get; private set; }
```
hishamco: I think readonly property in here is preferable

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
hishamco: This will be modify to
```csharp
resourcePath = resourceLocationAttribute?.ResourceFolder;
```
if we throw the exception above in the `ResourceLocationAttribute`

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -96,11 +112,23 @@ public IStringLocalizer Create(string baseName, string location)
...
+                var assembly = Assembly.Load(assemblyName);
+
+                var resourceLocation = _resourcesRelativePath;
+
+                // in another assembly
+                if (!string.Equals(assemblyName.Name, _hostingEnvironment.ApplicationName, StringComparison.Ordinal))
+                {
+                    var resourcesLocationAttribute = assembly.GetCustomAttribute<ResourceLocationAttribute>();
+                    resourceLocation = !string.IsNullOrEmpty(resourcesLocationAttribute?.ResourceFolder)
+                            ? resourceLocation = resourcesLocationAttribute.ResourceFolder + "."
```
hishamco: Again we can avoid the `string.IsNullOrEmpty` check everywhere

dougbu:
@ryanbrandenburg I've read some related comments elsewhere in this repo or perhaps Home or MVC. Could you link to the relevant issues or comments to help me w/ the details? In particular what is the problem this design PR is aimed to solve?

hishamco:
@dougbu AFAIK this design related to the issues https://github.com/aspnet/Localization/issues/281#issuecomment-247173762, https://github.com/aspnet/Localization/issues/275#issuecomment-247174158

test/ResourcesClassLibraryNoAttribute/project.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+  "version": "1.0.0-*",
+  "buildOptions": {
+    "warningsAsErrors": true,
+    "keyFile": "../../tools/Key.snk"
+  },
+  "dependencies": {
+    "NETStandard.Library": "1.6.0"
```
dougbu: Why not use a lower .NET standard library?

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
```diff
@@ -12,7 +12,27 @@ namespace Microsoft.AspNetCore.Localization.FunctionalTests
     public class LocalizationTest
     {
         private static readonly string _applicationPath = Path.Combine("test", "LocalizationWebsite");
-        
+
+        [ConditionalTheory]
+        [OSSkipCondition(OperatingSystems.Linux)]
+        [OSSkipCondition(OperatingSystems.MacOSX)]
```
ryanbrandenburg: So far as I can tell the idea is to avoid running CLR on Non-windows. But there seems to be a lot of duplication, I can file an issue to clean all of the tests in this format up at once.

ryanbrandenburg:
:up::date:

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
dougbu: plz

src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+                throw new ArgumentNullException(nameof(resourceLocation));
+            }
+
+            ResourceLocation = resourceLocation;
+        }
+
+        /// <summary>
+        /// The location of the resources for this Assembly
+        /// </summary>
+        public readonly string ResourceLocation;
```
hishamco: Readonly property not a ```readonly``` field
```csharp
public string ResourceLocation { get; }
```

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -170,6 +174,20 @@ public IStringLocalizer Create(string baseName, string location)
...
         }
 
+        protected virtual string GetResourcePrefix(string location, string baseName, string resourceLocation)
+        {
+            // Re-root the base name if a resources path is set
+            return location + "." + resourceLocation + TrimPrefix(baseName, location + ".");
+        }
+        private string GetResourcePath(Assembly assembly)
+        {
+            var resourceLocationAttribute = assembly.GetCustomAttribute<ResourceLocationAttribute>();
```
dougbu: To allow unit testing, need a method or two which has a `ResourceLocationAttribute` parameter and is available from test assembly. For example, confirm correct resource path is chosen if `ResourceLocation` contains directory separators.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -170,6 +174,20 @@ public IStringLocalizer Create(string baseName, string location)
...
+            // If we don't have an attribute assume all assemblies use the same resource location.
+            return string.IsNullOrEmpty(resourceLocationAttribute?.ResourceLocation)
```
dougbu: `ResourceLocation` can't be `null` if the attribute exists. Just check `resourceLocationAttribute == null`.

test/LocalizationWebsite/StartupResourcesInFolder.cs
```diff
@@ -43,6 +44,8 @@ public void ConfigureServices(IServiceCollection services)
             });
 
             var stringLocalizer = stringLocalizerFactory.Create("Test", location: null);
+            var assembly = typeof(StartupResourcesInFolder).GetTypeInfo().Assembly;
+            var stringLocalizerExpliciteLocation = stringLocalizerFactory.Create("Test", assembly.FullName);
```
dougbu: Frequente typoe in "Explicit"

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -109,5 +109,26 @@ public void Create_FromNameLocation_ReturnsNewResultForDifferentLocation()
...
+            var locOptions = new LocalizationOptions();
+            var options = new Mock<IOptions<LocalizationOptions>>();
+            options.Setup(o => o.Value).Returns(locOptions);
+            var factory = new ResourceManagerStringLocalizerFactory(hostingEnvironment.Object, localizationOptions: options.Object);
+            var location1 = typeof(ResourceManagerStringLocalizer).GetTypeInfo().Assembly.FullName;
+            var location2 = typeof(ResourceManagerStringLocalizerFactoryTest).GetTypeInfo().Assembly.FullName;
+
+            // Act
+            var result1 = factory.Create("baseName", null);
+            var result2 = factory.Create("baseName", null);
```
dougbu: Names for the `null` arguments

src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Gives the location of the resources for an Assembly.
+    /// </summary>
+    public class ResourceLocationAttribute : Attribute
```
dougbu: Be explicit about the `[AttributeUsage]`

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -157,11 +159,13 @@ public IStringLocalizer Create(string baseName, string location)
 
             location = location ?? _hostingEnvironment.ApplicationName;
 
-            baseName = GetResourcePrefix(baseName, location);
```
dougbu: Not using this `protected virtual` method is a breaking change. Those who have overridden it will no longer see their preferred behaviour.

src/Microsoft.Extensions.Localization/project.json
```diff
@@ -22,7 +22,8 @@
     "Microsoft.AspNetCore.Hosting.Abstractions": "1.1.0-*",
     "Microsoft.Extensions.DependencyInjection.Abstractions": "1.1.0-*",
     "Microsoft.Extensions.Localization.Abstractions": "1.1.0-*",
-    "Microsoft.Extensions.Options": "1.1.0-*"
+    "Microsoft.Extensions.Options": "1.1.0-*",
+    "System.Reflection.Extensions": "4.0.1"
```
dougbu: `"4.0.1-*"`

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Logging;
+
+namespace LocalizationWebsite
+{
+    public class StartupResourcesInClassLibrary
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddLocalization(options => options.ResourcesPath = "Resources");
```
dougbu: Separate from work in this design PR but do we thoroughly test use of `ResourcesPath` e.g. values containing directory separators?

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
```diff
@@ -0,0 +1,69 @@
...
+            var noAttributeName = new AssemblyName(noAttributeAssembly.FullName).Name;
+            var noAttributeNameStringLocalizer = stringLocalizerFactory.Create(
+                nameof(ResourcesClassLibraryNoAttribute.Model),
+                noAttributeName);
+
+            var withAttributeAssembly = typeof(ResourcesClassLibraryWithAttribute.Model).GetTypeInfo().Assembly;
+            var withAttributeName = new AssemblyName(withAttributeAssembly.FullName).Name;
+            var withAttributeNameStringLocalizer = stringLocalizerFactory.Create(
+                nameof(ResourcesClassLibraryWithAttribute.Model),
+                withAttributeName);
```
dougbu: 1. Only testing one `Create()` overload.
2. Is all this boilerplate really needed? Doesn't look particularly usable.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -99,8 +99,8 @@ public void Create_FromNameLocation_ReturnsNewResultForDifferentLocation()
             var options = new Mock<IOptions<LocalizationOptions>>();
             options.Setup(o => o.Value).Returns(locOptions);
             var factory = new ResourceManagerStringLocalizerFactory(hostingEnvironment.Object, localizationOptions: options.Object);
-            var location1 = typeof(ResourceManagerStringLocalizer).GetTypeInfo().Assembly.FullName;
-            var location2 = typeof(ResourceManagerStringLocalizerFactoryTest).GetTypeInfo().Assembly.FullName;
+            var location1 = new AssemblyName(typeof(ResourceManagerStringLocalizer).GetTypeInfo().Assembly.FullName).Name;
+            var location2 = new AssemblyName(typeof(ResourceManagerStringLocalizerFactoryTest).GetTypeInfo().Assembly.FullName).Name;
```
dougbu: Why change this test? Have the semantics w/o a `[ResourceLocation]` changed?

ryanbrandenburg: The test gave the impression that we should be passing in the full assemblyName to `Create(string, string)` which would make the ResourceManager look for something like `TestAssembly, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e.Resources.TestClass`. We actually want to pass the "short" assembly name there. But we didn't notice because the Assert in that test only cares if it produces Same/NotSame factories.

TL:DR: The semantics haven't changed, this was just wrong to start with.

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
ryanbrandenburg: I agree that `IStringLocalizerFactory.Create(string, string)` requires an awful lot of boilerplate, and is confusing as compared to `IStringLocalizerFactory.Create(Type)`. I would recommend very heavily that people use the Type overload, but I don't know if we're to the point of deprecating it, and I don't know of any way to make it simpler to use.

ryanbrandenburg:
:up::date:

src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Gives the location of the resources for an Assembly.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false)]
```
dougbu: Be explicit about `Inherited` choice as well.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -62,7 +62,7 @@ protected virtual string GetResourcePrefix(TypeInfo typeInfo)
                 throw new ArgumentNullException(nameof(typeInfo));
             }
 
-            return GetResourcePrefix(typeInfo, _hostingEnvironment.ApplicationName, _resourcesRelativePath);
+            return GetResourcePrefix(typeInfo, new AssemblyName(typeInfo.Assembly.FullName).Name, _resourcesRelativePath);
```
dougbu: This change won't matter (`GetResourcePrefix()` will still return `typeInfo.FullName` which is a strange resource prefix) when `_resourcesRelativePath` is `null` or empty. If that's a possible scenario, should think the semantics through a bit more. If it's not possible, remove the ternary at [line 91](https://github.com/aspnet/Localization/pull/283/files#diff-ec3f46a4394df7156f745d976f053127R91).

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -155,21 +159,50 @@ public IStringLocalizer Create(string baseName, string location)
...
+                _resourceNamesCache);
+        }
+
+        protected virtual string GetResourcePrefix(string location, string baseName, string resourceLocation)
+        {
+            // Re-root the base name if a resources path is set
+            return location + "." + resourceLocation + TrimPrefix(baseName, location + ".");
+        }
+
+        protected virtual ResourceLocationAttribute GetResourceLocationAttribute(Assembly assembly)
```
dougbu: Not sure about the other `protected virtual` methods but please add a comment identifying this one as `virtual` for testing. Can the other two be `private` or (better) `private static`?

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
dougbu: ??

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -1,17 +1,102 @@
...
 
-namespace Microsoft.Extensions.Localization.Test
+namespace Microsoft.Extensions.Localization.Tests
 {
+    public class TestResourceManagerStringLocalizerFactory : ResourceManagerStringLocalizerFactory
+    {
+        private ResourceLocationAttribute _resourceLocationAttribute;
+
+        public Assembly Assembly { get; set; }
+        public string BaseName { get; set; }
```
dougbu: Curious: Why are these `public`ly settable?

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -109,5 +217,48 @@ public void Create_FromNameLocation_ReturnsNewResultForDifferentLocation()
             // Assert
             Assert.NotSame(result1, result2);
         }
+
+        [Fact]
+        public void Create_FromNameLocation_ResourcesPathDirectorySeperatorToDot()
```
dougbu: Plze spele separator correctly

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -109,5 +217,48 @@ public void Create_FromNameLocation_ReturnsNewResultForDifferentLocation()
...
+            var factory = new TestResourceManagerStringLocalizerFactory(
+                hostingEnvironment.Object,
+                options.Object,
+                resourceLocationAttribute: null);
+
+            // Act
+            var result1 = factory.Create("baseName", location: null);
+
+            // Assert
+            Assert.DoesNotContain(Path.DirectorySeparatorChar, factory.BaseName);
```
dougbu: Look for the actual expected resource path to match the test name. Should not be OS-sensitive.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -109,5 +217,48 @@ public void Create_FromNameLocation_ReturnsNewResultForDifferentLocation()
...
+        {
+            // Arrange
+            var hostingEnvironment = new Mock<IHostingEnvironment>();
+            hostingEnvironment.SetupGet(a => a.ApplicationName).Returns("Microsoft.Extensions.Localization.Tests");
+            var locOptions = new LocalizationOptions();
+            var options = new Mock<IOptions<LocalizationOptions>>();
+            options.Setup(o => o.Value).Returns(locOptions);
+            var factory = new ResourceManagerStringLocalizerFactory(hostingEnvironment.Object, localizationOptions: options.Object);
+            var location1 = typeof(ResourceManagerStringLocalizer).GetTypeInfo().Assembly.FullName;
+            var location2 = typeof(ResourceManagerStringLocalizerFactoryTest).GetTypeInfo().Assembly.FullName;
```
dougbu: What are `locationX` for?

test/ResourcesClassLibraryNoAttribute/project.json
dougbu: ??

dougbu:
:watch: for you to declare this non-design. Looks close to me but double-check the doc comments and extra / missing blank lines.

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
ryanbrandenburg: We do not, which I know because it was broken. I added Create_FromType_ResourcesPathDirectorySeperatorToDot and Create_FromNameLocation_ResourcesPathDirectorySeperatorToDot to cover this area.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: It is possible and correct for GetResourcePrefix to return typeInfo.FullName. In the case that we don't have a Resource path it would return something like `Microsoft.Extensions.Localization.Tests.ResourceManagerStringLocalizerFactoryTest` which is how it will be stored in the assembly. Or are you asking something else?

test/ResourcesClassLibraryNoAttribute/project.json
ryanbrandenburg: Because if I do I get a warning that it ended up with NETStandard.Library 1.6.0 anyway. What are the downside of using 1.6 considering that this a a test library?

dougbu: :ok: guess I was mostly curious why the project needed this particular .NET standard.

src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Gives the location of the resources for an Assembly.
```
dougbu: Gives -> Provides and nuke the second "the".

dougbu: Same "the" removal twice below.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
dougbu: :ok:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -155,21 +157,70 @@ public IStringLocalizer Create(string baseName, string location)
...
+                baseName = GetResourcePrefix(baseName, location);
+
+                return CreateIStringLocalizer(assembly, baseName);
             });
         }
 
+        /// <summary>
+        /// This method is virtual for testing purposes only.
+        /// </summary>
+        /// <param name="assembly">The assembly to create an IStringLocalizer for.</param>
```
dougbu: `<see cref="IStringLocalizer"/>`

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -155,21 +157,70 @@ public IStringLocalizer Create(string baseName, string location)
...
+        /// <param name="baseName">The base name of the resource to search for.</param>
+        /// <returns></returns>
```
dougbu: Add a blurb here and in the two empty `<returns>` elements below.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -155,21 +157,70 @@ public IStringLocalizer Create(string baseName, string location)
...
+                var assemblyName = new AssemblyName(location);
+                var assembly = Assembly.Load(assemblyName);
+                baseName = GetResourcePrefix(baseName, location);
+
+                return CreateIStringLocalizer(assembly, baseName);
             });
         }
 
+        /// <summary>
+        /// This method is virtual for testing purposes only.
```
dougbu: A fine `<remarks>` but what does this method _do_?

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -155,21 +157,70 @@ public IStringLocalizer Create(string baseName, string location)
...
+        /// <returns></returns>
+        protected virtual ResourceLocationAttribute GetResourceLocationAttribute(Assembly assembly)
+        {
+            return assembly.GetCustomAttribute<ResourceLocationAttribute>();
+        }
+
+        private string GetResourcePath(Assembly assembly)
+        {
+            var resourceLocationAttribute = GetResourceLocationAttribute(assembly);
+            // If we don't have an attribute assume all assemblies use the same resource location.
```
dougbu: Blank line above to improve readability.

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
```
dougbu: ©

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
dougbu: Where is `IStringLocalizerFactory.Create(Type)` tested?

test/ResourcesClassLibraryNoAttribute/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Reflection;
```
dougbu: :copyright:

test/ResourcesClassLibraryWithAttribute/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Reflection;
```
dougbu: :copyright:

dougbu:
:watch: for some doc comment cleanup

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
ryanbrandenburg: We create `IStringLocalizer`'s with that overload above this on L43 and L44, then use them on L62 and L66. They also have unit tests in ResourceManagerStringLocalizerFactoryTest.

ryanbrandenburg:
:up::date:

test/LocalizationWebsite/StartupResourcesInClassLibrary.cs
dougbu: :ok_hand:

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -1,17 +1,79 @@
...
 
+using System.IO;
 using System.Reflection;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Options;
 using Moq;
 using Xunit;
 
-namespace Microsoft.Extensions.Localization.Test
+namespace Microsoft.Extensions.Localization.Tests
```
dougbu: nit: Shouldn't this be in the `Microsoft.Extensions.Localization` namespace?

ryanbrandenburg: Every other file in this repo uses adds Tests to the namespace it's testing, so if that is the standard, we're breaking it uniformly here.

dougbu: Right, we're breaking the [guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#unit-test-class-naming) uniformly:
> Test class names end with Test and live in the same namespace as the class being tested.

But you're also right that this is a separate issue from this PR.

PR #287, opened 2016-09-23
Use existing CultureInfo

### LINKED ISSUE #203: Custom Cultures not working with RequestLocalizationMiddleware
### Title
Custom Cultures not working with `RequestLocalizationMiddleware`

### Functional impact
It is impossible to use custom settings for Cultures when using `RequestLocalizationMiddleware`

### Minimal repro steps
- Create custom CultureProvider
```
public class SampleCultureProvider : RequestCultureProvider
{
    public override Task<ProviderCultureResult> DetermineProviderCultureResult(HttpContext httpContext)
    {
        var providerResultCulture = new ProviderCultureResult("en-US");
        return Task.FromResult(providerResultCulture);
    }
}
```
- Create custom localization options (changed CurrencySymbol in this example)
```
private RequestLocalizationOptions CreateLocalizationOptions()
{
    var supportedCulturesList = new List<CultureInfo>
    {
        new CultureInfo("en-US", true) { NumberFormat = { CurrencySymbol = "kr" } }
    };

    var localizationOptions = new RequestLocalizationOptions
    {
        SupportedCultures = supportedCulturesList,
        SupportedUICultures = supportedCulturesList
    };
            
    localizationOptions.RequestCultureProviders.Clear();
    localizationOptions.RequestCultureProviders.Add(new SampleCultureProvider());
    return localizationOptions;
}
```
- Add RequestLocalizationMiddleware using these options
```
public void Configure(IApplicationBuilder app)
{
    app.UseIISPlatformHandler();

    var globalizationOptions = this.CreateLocalizationOptions();
    app.UseRequestLocalization(globalizationOptions, new RequestCulture(new CultureInfo("en-GB")));

    (...)
}
```
- display text using currency formatting
```
public void Configure(IApplicationBuilder app)
{
    (...)
    app.Run(async (context) =>
    {
        await context.Response.WriteAsync(123.ToString("C"));
    });
}
```
Sample solution reproducing this bug attached: [RequestLocalizationMiddlewareBug.zip](https://github.com/aspnet/Localization/files/98886/RequestLocalizationMiddlewareBug.zip)

### Expected result
Changed currency symbol should be displayed: **kr123.00**

### Actual result
Default currency symbol is used: **$123.00**

### Further technical details
Problem is - most likely - caused by `CultureInfoCache`. When `GetCultureInfo` is called `supportedCultures` are checked to see whether requested culture is supported. 

```
if (name == null || supportedCultures == null ||
    !supportedCultures.Any(supportedCulture => string.Equals(supportedCulture.Name, name, StringComparison.OrdinalIgnoreCase)))
{
    return null;
}
```
Later on, however, instead of adding provided culture to the cache new culture is created with a given name and added
```
var entry = _cache.GetOrAdd(name, n =>
{
    try
    {
        return new CacheEntry(CultureInfo.ReadOnly(new CultureInfo(n)));
    }
    catch (CultureNotFoundException)
    {
        (...)
    }
});
```
I guess that using
```
var culture = supportedCultures.Single(supportedCulture => string.Equals(supportedCulture.Name, name, StringComparison.OrdinalIgnoreCase));
return new CacheEntry(CultureInfo.ReadOnly(culture));

```
instead of
```
return new CacheEntry(CultureInfo.ReadOnly(new CultureInfo(n)));
```
should fix this

ryanbrandenburg:
Fixes #203. Given this fix `CultureInfoCache` could just as a easily be a helper function on the middleware since it's no longer an actual cache, but I wasn't sure if we would be able to get rid of the whole class like that. @DamianEdwards would that be possible?

### DIFF
--- a/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs
@@ -4,11 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Globalization;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Localization
@@ -157,13 +157,32 @@ namespace Microsoft.AspNetCore.Localization
             return null;
         }
 
+        private static CultureInfo GetCultureInfo(string name, IList<CultureInfo> supportedCultures)
+        {
+            // Allow only known culture names as this API is called with input from users (HTTP requests) and
+            // creating CultureInfo objects is expensive and we don't want it to throw either.
+            if (name == null || supportedCultures == null)
+            {
+                return null;
+            }
+            var culture = supportedCultures.FirstOrDefault(
+                supportedCulture => string.Equals(supportedCulture.Name, name, StringComparison.OrdinalIgnoreCase));
+
+            if (culture == null)
+            {
+                return null;
+            }
+
+            return CultureInfo.ReadOnly(culture);
+        }
+
         private static CultureInfo GetCultureInfo(
             string cultureName,
             IList<CultureInfo> supportedCultures,
             bool fallbackToParentCultures,
             int currentDepth)
         {
-            var culture = CultureInfoCache.GetCultureInfo(cultureName, supportedCultures);
+            var culture = GetCultureInfo(cultureName, supportedCultures);
 
             if (culture == null && fallbackToParentCultures && currentDepth < MaxCultureFallbackDepth)
             {
--- a/src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
+++ b/src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Extensions.Globalization
     /// <summary>
     /// Provides read-only cached instances of <see cref="CultureInfo"/>.
     /// </summary>
+    [Obsolete("This type is obsolete and will be removed in a future version.")]
     public static class CultureInfoCache
     {
         private static readonly ConcurrentDictionary<string, CacheEntry> _cache = new ConcurrentDictionary<string, CacheEntry>();
@@ -23,7 +24,7 @@ namespace Microsoft.Extensions.Globalization
         /// <param name="name">The culture name.</param>
         /// <param name="supportedCultures">The cultures supported by the application.</param>
         /// <returns>
-        /// A read-only cached <see cref="CultureInfo"/> or <c>null</c> a match wasn't found in
+        /// A read-only cached <see cref="CultureInfo"/> or <c>null</c> if a match wasn't found in
         /// <paramref name="supportedCultures"/>.
         /// </returns>
         public static CultureInfo GetCultureInfo(string name, IList<CultureInfo> supportedCultures)
--- a/test/LocalizationWebsite/StartupCustomCulturePreserved.cs
+++ b/test/LocalizationWebsite/StartupCustomCulturePreserved.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System.Collections.Generic;
+using System.Globalization;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Localization;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace LocalizationWebsite
+{
+    public class StartupCustomCulturePreserved
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddLocalization();
+        }
+
+        public void Configure(
+            IApplicationBuilder app)
+        {
+            app.UseRequestLocalization(new RequestLocalizationOptions
+            {
+                DefaultRequestCulture = new RequestCulture("en-US"),
+                SupportedCultures = new List<CultureInfo>()
+                {
+                    new CultureInfo("en-US") { NumberFormat= { CurrencySymbol = "kr" } }
+                },
+                SupportedUICultures = new List<CultureInfo>()
+                {
+                    new CultureInfo("en-US") { NumberFormat= { CurrencySymbol = "kr" } }
+                }
+            });
+
+            app.Run(async (context) =>
+            {
+                await context.Response.WriteAsync(10.ToString("C"));
+            });
+        }
+    }
+}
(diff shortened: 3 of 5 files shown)


src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
```diff
@@ -14,8 +13,6 @@ namespace Microsoft.Extensions.Globalization
     /// </summary>
     public static class CultureInfoCache
```
pranavkm: public API, so you can't remove it

src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
```diff
@@ -30,38 +27,18 @@ public static CultureInfo GetCultureInfo(string name, IList<CultureInfo> support
...
-                }
-            });
-
-            return entry.CultureInfo;
-        }
-
-        private class CacheEntry
-        {
-            public CacheEntry(CultureInfo cultureInfo)
+            var culture = supportedCultures.SingleOrDefault(
```
pranavkm: `FirstOrDefault`

test/Microsoft.Extensions.Globalization.CultureInfoCache.Tests/CultureInfoCacheTest.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.Extensions.Globalization
+{
+    public class CultureInfoCacheTest
+    {
+        [Fact]
+        public void CultureInfoCache_ReturnsModifiedCultureInfo()
+        {
+            // Arrange
+            var supportedCultures = new List<CultureInfo> {
```
pranavkm: `{` on the next line.

src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
```diff
@@ -30,38 +27,18 @@ public static CultureInfo GetCultureInfo(string name, IList<CultureInfo> support
         {
             // Allow only known culture names as this API is called with input from users (HTTP requests) and
             // creating CultureInfo objects is expensive and we don't want it to throw either.
-            if (name == null || supportedCultures == null ||
```
pranavkm: Could you modify the `<returns>` on this method to say ...`or <null> if a match` (the if is missing)

ryanbrandenburg:
:up::date: @DamianEdwards is OK with obsoleting it for now and just not using it anywhere. I've built that in now.

Localization.sln
```diff
@@ -31,9 +31,9 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "LocalizationWebsite", "test
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNetCore.Localization.FunctionalTests", "test\Microsoft.AspNetCore.Localization.FunctionalTests\Microsoft.AspNetCore.Localization.FunctionalTests.xproj", "{B1B441BA-3AC8-49F8-850D-E5A178E77DE2}"
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "ResourcesClassLibraryWithAttribute", "test\ResourcesClassLibraryWithAttribute\ResourcesClassLibraryWithAttribute.xproj", "{F27639B9-913E-43AF-9D64-BBD98D9A420A}"
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "ResourcesClassLibraryNoAttribute", "test\ResourcesClassLibraryNoAttribute\ResourcesClassLibraryNoAttribute.xproj", "{939FFC22-82BD-47B3-9F6C-89C62AD0A0F7}"
```
hishamco: Is this change a part of the previous PR #283?!! correct me if I'm wrong

src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs
```diff
@@ -157,13 +157,31 @@ private static void SetCurrentThreadCulture(RequestCulture requestCulture)
...
+        {
+            // Allow only known culture names as this API is called with input from users (HTTP requests) and
+            // creating CultureInfo objects is expensive and we don't want it to throw either.
+            if (name == null || supportedCultures == null)
+            {
+                return null;
+            }
+            var culture = supportedCultures.FirstOrDefault(
+                supportedCulture => string.Equals(supportedCulture.Name, name, StringComparison.OrdinalIgnoreCase));
+            if (culture == null)
```
hishamco: Please add a new line here

src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Extensions.Globalization
     /// <summary>
     /// Provides read-only cached instances of <see cref="CultureInfo"/>.
     /// </summary>
+    [Obsolete("This type is obsolete and will be removed in a future version.")]
```
hishamco: :+1:

test/ResourcesClassLibraryNoAttribute/ResourcesClassLibraryNoAttribute.xproj
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
hishamco: Same for the first comment

test/ResourcesClassLibraryWithAttribute/ResourcesClassLibraryWithAttribute.xproj
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
hishamco: Same for the first comment

DamianEdwards:
@ryanbrandenburg as discussed, is any change required to the middleware to ensure it uses the actual supplied instance of `CultureInfo` specified as being the default?

ajaybhargavb:
:shipit: after removing all the unrelated changes.

ryanbrandenburg:
@DamianEdwards it appears we already used the existing instance of `CultureInfo` for `DefaultRequestCulture` so no change needed there.

Localization.sln
ryanbrandenburg: The SLN mearly got jumbled in my rebase. The xproj items got accidentally included and will be removed.

hishamco: I see

src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Extensions.Globalization
...
     public static class CultureInfoCache
```
hishamco: Do we really need this class anymore? or is it there for backward compatibility?

hishamco:
Can we have a unit test, to make sure that's fix #203?

src/Microsoft.Extensions.Globalization.CultureInfoCache/CultureInfoCache.cs
DamianEdwards: We can't remove public API (including types) in minor releases, so we have to keep it around until v2.

PR #291, opened 2016-09-30
Organize Functional tests

### LINKED ISSUE #284: Functional tests have duplicate tests for OSSkipConditions
The Localization functional tests use OSSkipCondition to skip CLR on Linux and Mac, but the tests seem to run the CoreCLR version twice, which confuses the issue. Let's clean up the skip conditions and duplicate tests where possible.

ryanbrandenburg:
Fixes #284. Standardize and de-duplicate functional tests.

### DIFF
--- a/test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationSampleTest.cs
+++ b/test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationSampleTest.cs
@@ -11,36 +11,30 @@ namespace Microsoft.AspNetCore.Localization.FunctionalTests
 {
     public class LocalizationSampleTest
     {
-        private static readonly string _applicationPath =  Path.Combine("samples", "LocalizationSample");
+        private static readonly string _applicationPath = Path.Combine("samples", "LocalizationSample");
 
-        [ConditionalTheory]
+        [ConditionalFact]
         [OSSkipCondition(OperatingSystems.Linux)]
         [OSSkipCondition(OperatingSystems.MacOSX)]
-        [InlineData(RuntimeFlavor.Clr, "http://localhost:5080/", RuntimeArchitecture.x64)]
-        [InlineData(RuntimeFlavor.CoreClr, "http://localhost:5081/", RuntimeArchitecture.x64)]
-        public Task RunSite_WindowsOnly(RuntimeFlavor runtimeFlavor, string applicationBaseUrl, RuntimeArchitecture runtimeArchitecture)
+        public Task RunSite_WindowsOnly()
         {
             var testRunner = new TestRunner(_applicationPath);
             return testRunner.RunTestAndVerifyResponseHeading(
-                runtimeFlavor,
-                runtimeArchitecture,
-                applicationBaseUrl,
+                RuntimeFlavor.Clr,
+                RuntimeArchitecture.x64,
+                "http://localhost:5080",
                 "My/Resources",
                 "fr-FR",
                 "<h1>Bonjour</h1>");
         }
-
-        [ConditionalTheory]
-        [OSSkipCondition(OperatingSystems.Windows)]
-        //[InlineData(RuntimeFlavor.Clr, "http://localhost:5080/", RuntimeArchitecture.x64)] // Disabled due to https://github.com/dotnet/corefx/issues/9012
-        [InlineData(RuntimeFlavor.CoreClr, "http://localhost:5081/", RuntimeArchitecture.x64)]
-        public Task RunSite_NonWindowsOnly(RuntimeFlavor runtimeFlavor, string applicationBaseUrl, RuntimeArchitecture runtimeArchitecture)
+        [Fact]
+        public Task RunSite_AnyOS()
         {
             var testRunner = new TestRunner(_applicationPath);
             return testRunner.RunTestAndVerifyResponseHeading(
-                runtimeFlavor,
-                runtimeArchitecture,
-                applicationBaseUrl,
+                RuntimeFlavor.CoreClr,
+                RuntimeArchitecture.x64,
+                "http://localhost:5081/",
                 "My/Resources",
                 "fr-FR",
                 "<h1>Bonjour</h1>");
--- a/test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
+++ b/test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Localization.FunctionalTests
         private static readonly string _applicationPath = Path.Combine("test", "LocalizationWebsite");
 
         [Fact]
-        public Task Localization_CustomCulture()
+        public Task Localization_CustomCulture_AllOS()
         {
             var testRunner = new TestRunner(_applicationPath);
             return testRunner.RunTestAndVerifyResponse(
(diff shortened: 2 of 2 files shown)


test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationSampleTest.cs
```diff
@@ -11,14 +11,11 @@ namespace Microsoft.AspNetCore.Localization.FunctionalTests
...
-        private static readonly string _applicationPath =  Path.Combine("samples", "LocalizationSample");
+        private static readonly string _applicationPath = Path.Combine("samples", "LocalizationSample");
 
-        [ConditionalTheory]
-        [OSSkipCondition(OperatingSystems.Linux)]
-        [OSSkipCondition(OperatingSystems.MacOSX)]
-        [InlineData(RuntimeFlavor.Clr, "http://localhost:5080/", RuntimeArchitecture.x64)]
+        [Theory]
         [InlineData(RuntimeFlavor.CoreClr, "http://localhost:5081/", RuntimeArchitecture.x64)]
-        public Task RunSite_WindowsOnly(RuntimeFlavor runtimeFlavor, string applicationBaseUrl, RuntimeArchitecture runtimeArchitecture)
```
ryanbrandenburg: Move WindowsOnly back on top.

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationSampleTest.cs
```diff
@@ -11,14 +11,11 @@ namespace Microsoft.AspNetCore.Localization.FunctionalTests
...
     {
-        private static readonly string _applicationPath =  Path.Combine("samples", "LocalizationSample");
+        private static readonly string _applicationPath = Path.Combine("samples", "LocalizationSample");
 
-        [ConditionalTheory]
-        [OSSkipCondition(OperatingSystems.Linux)]
-        [OSSkipCondition(OperatingSystems.MacOSX)]
-        [InlineData(RuntimeFlavor.Clr, "http://localhost:5080/", RuntimeArchitecture.x64)]
+        [Theory]
         [InlineData(RuntimeFlavor.CoreClr, "http://localhost:5081/", RuntimeArchitecture.x64)]
```
dougbu: `[Theory]`s that run once are just strange. Later in this PR you've moved from the more common `[Fact]` with hard-coded flavour, URL and architecture to this style. I suggest standardizing on the opposite i.e. favouring `[Fact]`s or `[ConditionalFact]`s where possible.

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
```diff
@@ -122,61 +115,37 @@ public Task Localization_CustomCulture()
                 "Hello Hello Hello Hello");
         }
 
-        [ConditionalFact]
```
dougbu: About this point, things get confusing. Exactly what was duplicated and how did you de-dupe?

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
```diff
@@ -122,61 +115,37 @@ public Task Localization_CustomCulture()
...
+            RuntimeArchitecture runtimeArchitecture)
         {
             var testRunner = new TestRunner(_applicationPath);
             return testRunner.RunTestAndVerifyResponse(
-                RuntimeFlavor.CoreClr,
-                RuntimeArchitecture.x64,
-                "http://localhost:5073/",
+                runtimeFlavor,
+                runtimeArchitecture,
+                applicationBaseUrl,
```
dougbu: Here's one of the `[ConditionalFact]` -> `[Fact]` conversions that went strange. Just do the simple thing.

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationSampleTest.cs
```diff
@@ -11,36 +11,30 @@ namespace Microsoft.AspNetCore.Localization.FunctionalTests
...
-        [InlineData(RuntimeFlavor.CoreClr, "http://localhost:5081/", RuntimeArchitecture.x64)]
-        public Task RunSite_WindowsOnly(RuntimeFlavor runtimeFlavor, string applicationBaseUrl, RuntimeArchitecture runtimeArchitecture)
+        public Task RunSite_WindowsOnly()
         {
             var testRunner = new TestRunner(_applicationPath);
             return testRunner.RunTestAndVerifyResponseHeading(
-                runtimeFlavor,
-                runtimeArchitecture,
-                applicationBaseUrl,
+                RuntimeFlavor.CoreClr,
```
dougbu: `RuntimeFlavor.Clr`

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
dougbu: Appears a few tests went away completely and some test pairs switched `RuntimeFlavor` order. But it's not clear.

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
```diff
@@ -122,72 +117,39 @@ public Task Localization_CustomCulture()
...
-                "http://localhost:5072",
-                "ResourcesInFolder",
-                "fr-FR",
-                "Bonjour from StartupResourcesInFolder Bonjour from Test in resources folder Bonjour from Customer in resources folder Bonjour from Test in resources folder");
-        }
-
-        [ConditionalFact]
-        [OSSkipCondition(OperatingSystems.Windows)]
-        [FrameworkSkipCondition(RuntimeFrameworks.CoreCLR)]
-        public Task Localization_ResourcesInFolder_ReturnLocalizedValue_WithCultureFallback_Clr()
```
ryanbrandenburg: This test was removed entirely because we already have a Windows and an AllOS version.

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
```diff
@@ -203,30 +165,15 @@ public Task Localization_ResourcesInFolder_ReturnLocalizedValue_WithCultureFallb
...
-                "http://localhost:5076",
-                "ResourcesAtRootFolder",
-                "fr-FR",
-                "Bonjour from StartupResourcesAtRootFolder Bonjour from Test in root folder Bonjour from Customer in Models folder");
-        }
-
-        [ConditionalFact]
-        [OSSkipCondition(OperatingSystems.Windows)]
-        [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
-        public Task Localization_ResourcesAtRootFolder_ReturnLocalizedValue_CoreCLR_NonWindows()
```
ryanbrandenburg: This test was removed entirely because we already have a Windows and an AllOS version.

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs
ryanbrandenburg: I would say paying attention to the original `RuntimeFlavor` order of these tests is not super valuable, because they were very muddled already. Take this test for example, as it was written it would only run if the OS was non-windows and the framework was CLR, meaning that it will never run.

ryanbrandenburg:
:up::date: As discussed, the diff is now more of a mess, but the final file should be more parse-able.

test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationSampleTest.cs
ryanbrandenburg: Did I address your concerns here? If not I'm not sure what the question is.

dougbu: Looks good.

dougbu:
Very minor: Might be worth using a few more different ports in `LocalizationTest` or at least using the existing ports more evenly. E.g. `:5070` is used a lot but `:5073` and `:5075` aren't used and `:5074` and `:5076` are used just once each.

PR #300, opened 2016-11-07
Remove unnecessary code in ResourceManagerStringLocalizer

hishamco:
After the PR #299 I notice there's some unnecessary lines of code, also there's an extra check from awhile [here](https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs#L48) because the `AssemblyWrapper` is already perform this check

/cc @Eilon @ryanbrandenburg


ryanbrandenburg:
This looks good except our standard on empty constructors is to not include an empty line. If you'll remove those two empty lines we'll take yours over mine.

hishamco:
I was planned to do this before, but I hesitate 😄 , I will do it now

hishamco:
@ryanbrandenburg hope it's fine now

ryanbrandenburg:
Thanks @hishamco!

PR #303, opened 2016-11-17
Fix LocalizeString docs

hishamco:
The docs for `ResourceNotFound` is little bit confusing, it should set to `true` if the string is missing and alternative string value was used

### DIFF
--- a/src/Microsoft.Extensions.Localization.Abstractions/LocalizedString.cs
+++ b/src/Microsoft.Extensions.Localization.Abstractions/LocalizedString.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Extensions.Localization
         /// </summary>
         /// <param name="name">The name of the string in the resource it was loaded from.</param>
         /// <param name="value">The actual string.</param>
-        /// <param name="resourceNotFound">Whether the string was found in a resource. Set this to <c>false</c> to indicate an alternate string value was used.</param>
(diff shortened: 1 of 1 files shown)


Eilon:
@ryanbrandenburg , all yours.

src/Microsoft.Extensions.Localization.Abstractions/LocalizedString.cs
```diff
@@ -25,7 +25,7 @@ public LocalizedString(string name, string value)
         /// </summary>
         /// <param name="name">The name of the string in the resource it was loaded from.</param>
         /// <param name="value">The actual string.</param>
-        /// <param name="resourceNotFound">Whether the string was found in a resource. Set this to <c>false</c> to indicate an alternate string value was used.</param>
+        /// <param name="resourceNotFound">Whether the string was not found in a resource. Set this to <c>true</c> to indicate an alternate string value was used.</param>
```
ryanbrandenburg: I think 
>Whether the string was found in a resource.

was fine as it was. Good catch on having the boolean backwards!

hishamco: May be, but it confusing little bit, because the actual param name is `resourceNotFound`

ryanbrandenburg: I asked a couple others and they agree with you, I guess this is more of a personal preference on my part rather than an actual english rule. You're all good to go.

PR #305, opened 2016-11-18
Add pluralization abstraction APIs

hishamco:
This address issue #17 which I asked for long time back

/cc @ryanbrandenburg

### DIFF
--- a/src/Microsoft.Extensions.Localization.Abstractions/IPluralizeStringLocalizer.cs
+++ b/src/Microsoft.Extensions.Localization.Abstractions/IPluralizeStringLocalizer.cs
@@ -1,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Represents a service that provides localized strings with pluralization support.
+    /// </summary>
+    public interface IPluralizeStringLocalizer : IStringLocalizer
+    {
+        /// <summary>
+        /// Gets the plural form for a given <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">The name of the string resource.</param>
+        /// <param name="count">The number of items you are trying to pluralize.</param>
+        /// <returns>The pluralized string resource as a <see cref="LocalizedString"/>.</returns>
+        LocalizedString Pluralize(string name, int count);
+
+        /// <summary>
+        /// Gets the pluralization rule for a given language.
+        /// </summary>
+        /// <param name="twoLetterISOLanguageName">The language that used to retrieve the plural forms for.</param>
+        /// <returns> The pluralization rule as <see cref="PluralizationRule"/>.</returns>
+        PluralizationRule GetPluralRule(string twoLetterISOLanguageName);
+    }
+}
--- a/src/Microsoft.Extensions.Localization.Abstractions/IPluralizedStringLocalizerOfT.cs
+++ b/src/Microsoft.Extensions.Localization.Abstractions/IPluralizedStringLocalizerOfT.cs
@@ -1,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Represents an <see cref="IPluralizeStringLocalizer"/> that provides pluralized strings for <typeparamref name="T"/>.
+    /// </summary>
+    /// <typeparam name="T">The <see cref="System.Type"/> to provide pluralized strings for.</typeparam>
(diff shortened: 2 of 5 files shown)


hishamco:
Indeed this a great feature for localization if we add it out of the box. Also I notice that @ryanbrandenburg working on `POLocalizer` which I already did few months back :), so it would be nice to have `POLocalizer` on top of **Pluralization APIs** while PO files supports pluralization

@Eilon your feedback?

Eilon:
@hishamco thanks for sending the PR. A bunch of folks are on vacation so we're going to meet in ~2 weeks to discuss this. Thank you for your patience!

hishamco:
I see .. so that why I notice that almost the guys I knew they are not active last few days 😄

Eilon:
Yup, it's Thanksgiving in the US, so everyone has Thu/Fri/Sat/Sun off, and a lot of folks take extra time off as well around this time! 😄

hishamco:
@ryanbrandenburg any feedback about this

hishamco:
@DamianEdwards any feedback regarding this

src/Microsoft.Extensions.Localization.Abstractions/IPluralizeStringLocalizer.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <param name="count">The number of items you are trying to pluralize.</param>
+        /// <returns>The pluralized string resource as a <see cref="LocalizedString"/>.</returns>
+        LocalizedString Pluralize(string name, int count);
+
+        /// <summary>
+        /// Gets the pluralization rule for a given language.
+        /// </summary>
+        /// <param name="twoLetterISOLanguageName">The language that used to retrieve the plural forms for.</param>
+        /// <returns> The pluralization rule as <see cref="PluralizationRule"/>.</returns>
+        PluralizationRule GetPluralRule(string twoLetterISOLanguageName);
```
ryanbrandenburg: I assume the reason for including `GetPluralRule` on the public interface would be to allow allow client-side frameworks to decide which pluralization to use, if that's the case I don't think that returning a C# delegate is a consumable way of doing it.

ryanbrandenburg (commented):
We're having a meeting soon to discuss pluralization in general, and I believe the IPluralizableStringLocalizer is on the docket. I'll be sure to bring this PR into the discussion.

src/Microsoft.Extensions.Localization.Abstractions/IPluralizeStringLocalizer.cs
hishamco: May be, there 're different ways to think about this, either we pass the number that decide which form should be selected, but what i'm looking for is a simple and elegant way that code reduce the efforts by passing the language rule once and the APIs will take the rest of the stuff to find the proper form, so that's why I provide a code snippet in the unit tests to prove the concept.
Also we may think about better option to provide the language pluralization rules without tedious and repetitive code, AFAIK PO files are one of the format that support the language rule itself as a part of the file itself to choose the appropriate form. Hope we hear the feedback from other folks

hishamco:
Hope to see the Pluralization a good piece of technology, which is something I demand it for more than year ago, also I did few samples to prove the concept. I'm sure this is will make our localization APIs special that the previous versions of the ASP.NET, also it open up a new localization sources other than resx files such as PO which is one of the famous localization formats for localization world wide

src/Microsoft.Extensions.Localization.Abstractions/IPluralizeStringLocalizer.cs
ryanbrandenburg: Yeah, PO files allow you to include a set of ternary conditions described [here](https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html) to define the plurality of your strings.

hishamco:
@Eilon I notice that this PR go to the backlog is that mean PR #309 will go to the backlog too, because `POStringLocalizer` depend on this

hishamco:
@Eilon is there a plan to support ```POStringLocalizer``` and Pluralization in 2.0.0?

Eilon:
The POStringLocalizer at this point is probably just a sample - we're not committed to shipping it as an official feature. However, it doesn't depend on supporting pluralization because not all PO files have pluralization. We might just implement a subset of PO file functionality.

hishamco:
While PO files support pluralization, i think it make sense to implement `IPluralizeStringLocalizer` instead of `IStringLocalizer` even not all PO files have pluralization. One more thing PO is one of the popular localization files that have been used for awhile, especially for non Windows OS, so it would be nice to have such `POStringLocalizer` using simple implementation, because the current proposal is complicated little bit

natemcmaster:
As you may already know, we're going to archive this repo. I'm working on moving the code today. Once complete, the Microsoft.Extensions.\* projects will be in https://github.com/aspnet/Extensions, and Microsoft.AspNetCore.\* will be in https://github.com/aspnet/AspNetCore. If this is something you'd like to get merged, please open a new PR against the new repos.

Thanks,
Nate

PR #309, opened 2016-12-14
[Design] POStringLocalizer

ryanbrandenburg:
This is a design PR for the POStringLocalizer work. Missing from the is the pluralization work (other than just to parse the file) because we haven't decided on what the shape of IPluralizedStringLocalizer is.


hishamco:
Finally we have it 😄 , great job @ryanbrandenburg I appreciate it.
One thing why we didn't finalize the pluralization stuff after that build this on top of that, as we know PO files support pluralization by its nature

hishamco:
@ryanbrandenburg can you fix the checks failure?

src/Microsoft.Extensions.Localization/Internal/POLines/ContextLine.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System;
+using System.Text;
+
+namespace Microsoft.Extensions.Localization.Internal.POLines
+{
+    public class ContextLine : TokenLine
+    {
+        public override string Token
```
hishamco:
```csharp
public override string Token => "msgctxt ";
```

src/Microsoft.Extensions.Localization/Internal/POLines/ContextLine.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            get
+            {
+                return "msgctxt ";
+            }
+        }
+
+        public override Line Parse(string value)
```
hishamco:
```csharp
public override Line Parse(string value) =>
    new ContextLine
    {
        Value = TrimToken(new StringBuilder(value))
    };
```

src/Microsoft.Extensions.Localization/Internal/POLines/FlagLine.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+using System.Text;
+using System.Linq;
+
+namespace Microsoft.Extensions.Localization.Internal.POLines
+{
+    public class FlagLine : CommentBase
+    {
+        public override string Token
```
hishamco:
```csharp
public override string Token => "#, ";
```

src/Microsoft.Extensions.Localization/Internal/POLines/FlagLine.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            get
+            {
+                return "#, ";
+            }
+        }
+
+        public override Line Parse(string value)
```
hishamco:
```csharp
public override Line Parse(string value) =>
    new FlagLine
    {
        Value = TrimToken(new StringBuilder(value)).ToString().Split(',').ToList()
    };
```

src/Microsoft.Extensions.Localization/Internal/POLines/Line.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                //TODO: Check for values at the end
+            }
+            else
+            {
+                throw new FormatException("Line malformed");
+            }
+        }
+
+        private bool IsQuote(char character)
```
hishamco:
```csharp
private bool IsQuote(char character) =>
    character == '"' || character == '\'';
```

src/Microsoft.Extensions.Localization/Internal/POLines/Line.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            return character == '"' || character == '\'';
+        }
+
+        private StringBuilder Unescape(StringBuilder value)
```
hishamco:
```csharp
private StringBuilder Unescape(StringBuilder value) =>
    value.Replace("\\\"", "\"").Replace("\\'", "'");
```

src/Microsoft.Extensions.Localization/Internal/POLines/LiteralLine.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System;
+using System.Text;
+
+namespace Microsoft.Extensions.Localization.Internal.POLines
+{
+    public class LiteralLine : Line
+    {
+        public override Line Parse(string value)
```
hishamco:
```csharp
public override Line Parse(string value) =>
    new LiteralLine
    {
        Value = TrimQuotes(new StringBuilder(value))
    };
```

hishamco (changes requested):

hishamco:
@ryanbrandenburg It will be nice if `POLocalizer` implement `IPluralizeStringLocalizer` that I suggest in PR #305 instead of `IStringLocalizer` because it doesn't support pluralization, so that's why you need to do it yourself in each implementation that may support pluralization. My aim to eliminate the tedious and repetitive code that we wrote in such cases.

Second thing the amount of code to parse PO is too much, can we simplify this like what we did in INI configuration as example?

hishamco:
@sebastienros is that mean that `POStringLocalizer` will not be supported?!!

sebastienros:
Fortunately no. It's just that this branch required more changes so it will be started fresh.

hishamco:
I see, furthermore I saw that implementation is complicated little bit, It would be nice if we can simplify parsing like what I have seen in INI Configuration

thanks

PR #316, opened 2017-01-30
Remove dependency on Hosting.Abstractions

ryanbrandenburg:
This removes the dependency on `Microsoft.AspNetCore.Hosting.Abstractions`.

### DIFF
--- a/src/Microsoft.Extensions.Localization/exceptions.net45.json
+++ b/src/Microsoft.Extensions.Localization/exceptions.net45.json
@@ -1,0 +1,9 @@
+[
+    {
+        "OldTypeId": "public class Microsoft.Extensions.Localization.ResourceManagerStringLocalizerFactory : Microsoft.Extensions.Localization.IStringLocalizerFactory",
+        "OldMemberId": "public .ctor(Microsoft.AspNetCore.Hosting.IHostingEnvironment hostingEnvironment, Microsoft.Extensions.Options.IOptions<Microsoft.Extensions.Localization.LocalizationOptions> localizationOptions)",
+        "NewTypeId": "public class Microsoft.Extensions.Localization.ResourceManagerStringLocalizerFactory : Microsoft.Extensions.Localization.IStringLocalizerFactory",
+        "NewMemberId": "public .ctor(Microsoft.Extensions.Options.IOptions<Microsoft.Extensions.Localization.LocalizationOptions> localizationOptions)",
+        "Kind": "Modification"
+    }
+]
--- a/test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
+++ b/test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved. 
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
 using System.Reflection;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Options;
 using Moq;
 using Xunit;
@@ -18,10 +18,9 @@ namespace Microsoft.Extensions.Localization.Tests
         public string BaseName { get; private set; }
 
         public TestResourceManagerStringLocalizerFactory(
-            IHostingEnvironment hostingEnvironment,
             IOptions<LocalizationOptions> localizationOptions,
             ResourceLocationAttribute resourceLocationAttribute)
-            : base(hostingEnvironment, localizationOptions)
+            : base(localizationOptions)
         {
             _resourceLocationAttribute = resourceLocationAttribute;
         }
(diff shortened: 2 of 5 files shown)


src/Microsoft.Extensions.Localization/IApplicationNameProvider.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.Extensions.Localization
```
JunTaoLuo: Copyright

src/Microsoft.Extensions.Localization/IApplicationNameProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Localization
+{
+    public interface IApplicationNameProvider
```
rynowak: This problem with this approach is that your `AddLocalization` methods no longer work out of the box. This removes the dependency on Hosting, but replaces it with an implicit dependency on something else. 

Why not add the application name to `LocalizationOptions` and do both of:
- define the behavior to something sane when application name isn't set
- teach MVC to set the application name for your when localization is in use

This doesn't require any new abstractions or plumbing, and has the advantage that localization can be used outside of MVC.

rynowak: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Localization/Internal/MvcLocalizationServices.cs

hishamco:
What if we didn't allow null-ability for `location` here https://github.com/aspnet/Localization/blob/07e774146306d7f838aea279655b9a6058e1c7a1/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs#L158 instead of https://github.com/aspnet/Localization/blob/07e774146306d7f838aea279655b9a6058e1c7a1/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs#L165

src/Microsoft.Extensions.Localization/IApplicationNameProvider.cs
Eilon: Hmm, I do wonder - which of the Entropy (or other?) samples need to be updated due to this? I admit I hadn't quite considered those implications.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved. 
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
Eilon: Needed?

ryanbrandenburg: Needed for when we `Assert.Throw<ArgumentNullException>`

hishamco: That's what I wanna say 😄

hishamco:
Seems the final changes looks fine 👍, well done!!

hishamco:
@ryanbrandenburg seems we need to remove these too https://github.com/aspnet/Localization/blob/6f3a68d80dda9681e2fc3ff0020ba38291d41aae/src/Microsoft.Extensions.Localization/baseline.netcore.json#L320-L323, https://github.com/aspnet/Localization/blob/6f3a68d80dda9681e2fc3ff0020ba38291d41aae/src/Microsoft.Extensions.Localization/baseline.net45.json#L320-L323

Eilon:
@hishamco we don't update the baselines; rather, we just add exclusions in a separate list.

hishamco:
Thanks, because I thought that may required for the docs, coz tell now I don't have any idea about these files 😄

src/Microsoft.Extensions.Localization/exceptions.net45.json
```diff
@@ -0,0 +1,9 @@
+[
+    {
+        "OldTypeId": "public class Microsoft.Extensions.Localization.ResourceManagerStringLocalizerFactory : Microsoft.Extensions.Localization.IStringLocalizerFactory",
+        "OldMemberId": "public .ctor(Microsoft.AspNetCore.Hosting.IHostingEnvironment hostingEnvironment, Microsoft.Extensions.Options.IOptions<Microsoft.Extensions.Localization.LocalizationOptions> localizationOptions)",
+        "NewTypeId": "public class Microsoft.Extensions.Localization.ResourceManagerStringLocalizerFactory : Microsoft.Extensions.Localization.IStringLocalizerFactory",
+        "NewMemberId": "public .ctor(Microsoft.Extensions.Options.IOptions<Microsoft.Extensions.Localization.LocalizationOptions> localizationOptions)",
```
Eilon: @javiercn - how exactly does the tool know that this ctor was "modified" as opposed to an old one got deleted, and a new one showed up? Is that not fragile?

javiercn: We have a modification kind

javiercn: You set it to Removal, Modification or Adition

ryanbrandenburg: @Eilon is that enough of an explanation and this is approved now?

Eilon: Yep.

hishamco:
The AppVeyor failed!!

ryanbrandenburg:
An announcement was made for this breaking change in aspnet/Announcements#218.

PR #323, opened 2017-02-05
Remove CultureInfoCache

hishamco:
After removing the dependency on Hosting.Abstractions in PR #316, I think it's the time to remove the `CultureInfoCache` which another breaking-change that should seen in **v2.0.0**

/cc @ryanbrandenburg @Eilon


hishamco:
React to VS 2017 upgrade

ryanbrandenburg:
@hishamco your tests are failing because you haven't removed the reference to the `CultureInfoCache` project from other projects. Running .\build.cmd should show you which ones are broken.

@Eilon 2.0.0 seems like a good time to get rid of this since it's already obsoleted, yeah?

hishamco:
This is was the issue https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNetCore.Localization/Microsoft.AspNetCore.Localization.csproj#L15, It was fine before VS 2017 upgrade, so that's why i ping you to accept this before .csproj upgrade 😄

Eilon:
Yup sounds good.

@ryanbrandenburg I assigned https://github.com/aspnet/Localization/issues/325 to you to track.

Issue #293, opened 2016-10-02
Localization always returns english clulture


ajtowf:
My views contains swedish text e.g.
@Localizer["Välkommen"] // means welcome

If I have a english resource for my view e.g. Index.en.resx this text will get translated to the english text no matter what! Even if I configure the app not to even support english, eg:

```
services.Configure<RequestLocalizationOptions>(options =>
        {
            var supportedCultures = new[]
            {
                new CultureInfo("sv"),
                //new CultureInfo("en")
            };

            options.DefaultRequestCulture = new RequestCulture(culture: "sv", uiCulture: "sv");
            options.SupportedCultures = supportedCultures;
            options.SupportedUICultures = supportedCultures;
            options.RequestCultureProviders.Insert(0, new CustomRequestCultureProvider(context =>
            {
                return Task.FromResult(new ProviderCultureResult("sv", "sv"));
            }));
        });
```

and in configure app:

```
var locOptions = app.ApplicationServices.GetService<IOptions<RequestLocalizationOptions>>();
        app.UseRequestLocalization(locOptions.Value);
```

Pretty frustrating, so the workaround for now is to simply rename the resource files to some other culture, norweigan for instance e.g. Index.no.resx.

hishamco:
I configure the localization options like this, but the configuration that you did should work too. To make things work as expected you should add a Swedish resource file like the following **Resources/Views/ControllerName/ViewName.sv.resx**

ajtowf:
What you're proposing is something entirely different. My views contains Swedish text so there is no .sv.resx files since Swedish is the default. The user may choose to switch to English however, and if there is any ViewName.**en**.resx files, the text will be localized to English no matter what. See the difference?

ajtowf:
My workaround for this is to put the English localization into some bogus language like Norwegian e.g. ViewName.**no**.resx and camouflage Norwegian as English.

hishamco:
@ajtowf can you update to the latest version, I tried to use the localization sample with English & Swedish as supported cultures, Swedish is the default one. I added Startup.en.resx, so when I visit `http://localhost:5000` it retrieve **"Välkommen"**, when I visit `http://localhost:5000/?culture=en&ui-culture=en' I got **"Welcome"**, these are the correct and expected result.
If i'm not wrong it was bug, because I saw a similar result like what you mentioned before I update to the latest changes

ryanbrandenburg:
[...]

PR #333, opened 2017-02-24
English does not override

ryanbrandenburg:
Add a test for #293.

### DIFF
--- a/test/LocalizationWebsite/LocalizationWebsite.csproj
+++ b/test/LocalizationWebsite/LocalizationWebsite.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <Import Project="..\..\build\common.props" />
 
--- a/test/LocalizationWebsite/StartupResourcesInFolder.cs
+++ b/test/LocalizationWebsite/StartupResourcesInFolder.cs
@@ -26,7 +26,10 @@ namespace LocalizationWebsite
             ILoggerFactory loggerFactory,
             IStringLocalizerFactory stringLocalizerFactory,
             IStringLocalizer<StartupResourcesInFolder> startupStringLocalizer,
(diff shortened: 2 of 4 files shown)


test/LocalizationWebsite/LocalizationWebsite.csproj
```diff
@@ -21,4 +21,10 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="1.2.0-*" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="Resources\StartupCustomCulturePreserved.en-US.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+    </EmbeddedResource>
+  </ItemGroup>
```
dougbu: Is this because no generator is wired up by default (wow, if so)? Or, does this resx need a specific generator?

test/LocalizationWebsite/StartupResourcesInFolder.cs
```diff
@@ -26,7 +26,8 @@ public void ConfigureServices(IServiceCollection services)
             ILoggerFactory loggerFactory,
             IStringLocalizerFactory stringLocalizerFactory,
             IStringLocalizer<StartupResourcesInFolder> startupStringLocalizer,
-            IStringLocalizer<Customer> custromerStringLocalizer)
+            IStringLocalizer<Customer> custromerStringLocalizer,
+            IStringLocalizer<StartupCustomCulturePreserved> programLocalizer)
```
dougbu: Suggest adding a comment referencing #293 and spelling out the scenario this line checks.

test/LocalizationWebsite/LocalizationWebsite.csproj
ryanbrandenburg: VS seems to be automatically adding these when you create a new resource file even though things work fine without them. I've given feedback on that through the in-VS mechanism and will keep a closer eye on it in the future.

PR #334, opened 2017-02-26
Remove RuntimeIdentifier from web projects

hishamco:
The runtime Ids should be removed from .csproj web projects while https://github.com/dotnet/sdk/issues/396 is got fixed


pranavkm:
Thanks @hishamco! FYI @natemcmaster \ @dougbu since you were asking about this in the Mvc PR.

natemcmaster:
FYI we'll be doing a general sweep of our repos to clean up workarounds such as this one.

hishamco:
@pranavkm @natemcmaster I can go through all the repos if you want

dougbu:
@hishamco I bet @natemcmaster or @pranavkm already have a script cued up to do this everywhere, including repos that aren't yet public.

In any case, we should old off removing all `RuntimeIdentifier` settings. It breaks testing in Visual Studio at least sometimes. Probably not an issue in this repo because the problems seem focused on the Mvc function tests and those are unusual.

hishamco:
Ok, all the best 👍

PR #348, opened 2017-03-22
Change compilation targets.

pranavkm:
* Remove net451 as a compilation target
* Upgrade to netcoreapp2.0

### DIFF
--- a/test/ResourcesClassLibraryNoAttribute/ResourcesClassLibraryNoAttribute.csproj
+++ b/test/ResourcesClassLibraryNoAttribute/ResourcesClassLibraryNoAttribute.csproj
@@ -3,8 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net452;netstandard1.6</TargetFrameworks>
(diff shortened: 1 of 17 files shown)


test/ResourcesClassLibraryNoAttribute/ResourcesClassLibraryNoAttribute.csproj
```diff
@@ -3,8 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net452;netstandard1.6</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard1.6</TargetFrameworks>
+    <TargetFramework>netstandard1.0</TargetFramework>
```
ryanbrandenburg: I assume that in these two cases where we went down some versions that the standard was "lowest one that works"?

pranavkm: yup

PR #363, opened 2017-04-28
React to Logging API changes

BrennanConroy:
Preview1 approved


BrennanConroy:
Planning on posting one once all the Logging stuff is in

PR #366, opened 2017-05-06
Migration

JunTaoLuo:


test/ResourcesClassLibraryNoAttribute/ResourcesClassLibraryNoAttribute.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.0</TargetFramework>
```
JunTaoLuo: these test packages don't really matter but for consistency never hurts.

pranavkm: I'd leave them be since it specifically says "Library"

PR #367, opened 2017-05-10
Remove unnecessary package references

pakrym:
https://github.com/aspnet/Coherence/issues/175


test/Microsoft.Extensions.Localization.Tests/Microsoft.Extensions.Localization.Tests.csproj
```diff
@@ -21,7 +21,6 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
```
pranavkm: Remove the node

PR #369, opened 2017-05-16
Add a builder API for configuring UseRequestLocalization

hishamco:
The PR #354 merged & closed accidentally during a mistake I did in my local, so I create this one instead


hishamco:
@ryanbrandenburg any plans to accepts this, because Damian talked about this long time back

samples/LocalizationSample/Startup.cs
```diff
@@ -23,31 +22,19 @@ public void ConfigureServices(IServiceCollection services)
...
-            //options.RequestCultureProviders.Insert(0, new CustomRequestCultureProvider(async context =>
-            //{
-
-            //}));
+            var supportedCultures = new [] { "en-US", "en-AU", "en-GB", "es-ES", "ja-JP", "fr-FR", "zh", "zh-CN" };
+            app.UseRequestLocalization(options =>
+                options
+                    .AddSupportedCultures(supportedCultures)
+                    .AddSupportedUICultures(supportedCultures)
+                    .SetDefaultCulture(supportedCultures[0])
```
mkArtakMSFT: Can we not take dependency through an index, as that will break if someone will modify the list in the future. Let's keep the 'en-US' as the default explicitly.

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
```diff
@@ -109,5 +109,53 @@ public RequestCulture DefaultRequestCulture
...
+            return this;
+        }
+
+        /// <summary>
+        /// Set the default culture to be used by the application when a supported culture could not be determined by
+        /// one of the configured <see cref="IRequestCultureProvider"/>s.
+        /// </summary>
+        /// <param name="defaultCulture">The default culture to be set.</param>
+        /// <returns>The <see cref="RequestLocalizationOptions"/>.</returns>
+        public RequestLocalizationOptions SetDefaultCulture(string defaultCulture)
```
mkArtakMSFT: What will happen, if the value of the defaultCulture is not listed in the supportedCultures?

samples/LocalizationSample/Startup.cs
```diff
@@ -23,31 +22,19 @@ public void ConfigureServices(IServiceCollection services)
...
-            };
-            // Optionally create an app-specific provider with just a delegate, e.g. look up user preference from DB.
-            // Inserting it as position 0 ensures it has priority over any of the default providers.
-            //options.RequestCultureProviders.Insert(0, new CustomRequestCultureProvider(async context =>
-            //{
-
-            //}));
+            var supportedCultures = new [] { "en-US", "en-AU", "en-GB", "es-ES", "ja-JP", "fr-FR", "zh", "zh-CN" };
+            app.UseRequestLocalization(options =>
+                options
```
rynowak: I realize that this is based on my own suggestion https://github.com/aspnet/Localization/issues/127 - (from 2.5 years ago wow!), but I'm going to suggest a modification on what I said.

One comment here for posterity - the design of how work with these kinds of builders has changed since I made the proposal back in the much simpler time of 2015.

---
The typical pattern is:

```
services.AddLocalization(...) // returns a builder - that builder will modify options
```

```
app.UseRequestLocalization(); // no args version gets options from services
```
---

Unfortunately, we already have a `services.AddLocalization()` and it returns `IServiceCollection`. Changing this would be a breaking change and we would not do that for minor API polish like this. So we've painted ourselves into a corner. 

I think this PR based on my original proposal is the best we can do. But I wanted to point out that it's not consistent with our other API shapes.

src/Microsoft.AspNetCore.Localization/ApplicationBuilderExtensions.cs
```diff
@@ -43,12 +43,77 @@ public static IApplicationBuilder UseRequestLocalization(this IApplicationBuilde
...
 
             return app.UseMiddleware<RequestLocalizationMiddleware>(Options.Create(options));
         }
+
+        /// <summary>
+        /// Adds the <see cref="RequestLocalizationMiddleware"/> to automatically set culture information for
+        /// requests based on information provided by the client. 
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/>.</param>
+        /// <param name="optionsAction"></param>
```
rynowak: This needs a description. It also needs to say that we're going to instantiate a **new** `RequestLocalizationOptions` that doesn't come from the services.

src/Microsoft.AspNetCore.Localization/ApplicationBuilderExtensions.cs
```diff
@@ -43,12 +43,77 @@ public static IApplicationBuilder UseRequestLocalization(this IApplicationBuilde
...
+            }
+
+            if (cultures == null)
+            {
+                throw new ArgumentNullException(nameof(cultures));
+            }
+
+            if (cultures.Length == 0)
+            {
+                throw new ArgumentException(nameof(cultures));
```
rynowak: This needs a message

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
```diff
@@ -109,5 +109,53 @@ public RequestCulture DefaultRequestCulture
         /// </list>
         /// </summary>
         public IList<IRequestCultureProvider> RequestCultureProviders { get; set; }
+
+        
+        /// <summary>
+        /// Adds the set of the supported cultures by the application.
```
rynowak: I think it's a good idea to add the links using `<see cref="CultureInfo.CurrentCulture"/>` and `<see cref="CultureInfo.CurrentUICulture"/>` to these. These two concepts are frequently misunderstood.

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
```diff
@@ -109,5 +109,53 @@ public RequestCulture DefaultRequestCulture
...
+            var supportedUICultures = new List<CultureInfo>();
+            foreach (var culture in uiCultures)
+            {
+                supportedUICultures.Add(new CultureInfo(culture));
+            }
+
+            SupportedUICultures = supportedUICultures;
+            return this;
+        }
+
```
rynowak: Your comment on the earlier bug included a method that added a list of cultures to both lists, any thoughts about that? Do you think that's important?

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
```diff
@@ -109,5 +109,53 @@ public RequestCulture DefaultRequestCulture
...
+            SupportedCultures = supportedCultures;
+            return this;
+        }
+
+        /// <summary>
+        /// Adds the set of the supported UI cultures by the application.
+        /// </summary>
+        /// <param name="uiCultures">The UI cultures to be added.</param>
+        /// <returns>The <see cref="RequestLocalizationOptions"/>.</returns>
+        public RequestLocalizationOptions AddSupportedUICultures(params string[] uiCultures)
```
rynowak: Should we have overloads that also accept `CultureInfo` ?

src/Microsoft.AspNetCore.Localization/ApplicationBuilderExtensions.cs
```diff
@@ -43,12 +43,77 @@ public static IApplicationBuilder UseRequestLocalization(this IApplicationBuilde
...
+            }
+
+            var options = new RequestLocalizationOptions()
+                .AddSupportedCultures(cultures)
+                .AddSupportedUICultures(cultures)
+                .SetDefaultCulture(cultures[0]);
```
ryanbrandenburg: I'm not in love with the idea of taking the first culture as the default. It seems undiscoverable/confusing. I would suggest just having a defaultCulture parameter even if it does lead to a little repetition. @rynowak what do you think?

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
```diff
@@ -109,5 +109,53 @@ public RequestCulture DefaultRequestCulture
...
+            {
+                supportedUICultures.Add(new CultureInfo(culture));
+            }
+
+            SupportedUICultures = supportedUICultures;
+            return this;
+        }
+
+        /// <summary>
+        /// Set the default culture to be used by the application when a supported culture could not be determined by
```
ryanbrandenburg: Nitpick on wording here, maybe `Set the DefaultCulture, which is used by the application` etc.

src/Microsoft.AspNetCore.Localization/ApplicationBuilderExtensions.cs
rynowak: It comes down to scenarios IMO - do you **always** want a default culture?

ryanbrandenburg: As it is now if you don't have a default culture set and a culture can't be determined from the request then it seems the culture the user receives will depend on the default THREAD culture, which I feel like is a scary and confusing fallback story. See [here](https://github.com/aspnet/Localization/blob/8b03ee8c81b61a852a586740c4db7b09c17f5e76/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs#L146).

rynowak: Which is a good reason to have a sensible default. What would the API look like with an additional parameter? 

I'd actually suggest baking that behavior into the middleware rather than the options/builder. it seems like **anyone** would agree that's better than the current behavior. Is there a historical reason for this?

samples/LocalizationSample/Startup.cs
hishamco: I don't think this will break while all the cultures are supported

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
hishamco: `Set the default culture which is used by the application` etc, may better as far as I saw in the rest of the docs

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
hishamco: It will fallback to the `en-US` if i'm not wrong, this is similar to setting the `DefaultRequestCulture` with invalid culture

IMHO it would be nice to write a unit test to cover this case

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
hishamco: I'm not sure why we need another overload while we already have `SupportedUICultures` property!!

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
hishamco: Which one you mean?

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
ryanbrandenburg: I'm also fine with that wording.

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
hishamco: Can you elaborate where should I reference them? Because the builder methods are using `SupportedCultures` & `SupportedUICultures`

src/Microsoft.AspNetCore.Localization/ApplicationBuilderExtensions.cs
```diff
@@ -43,12 +43,77 @@ public static IApplicationBuilder UseRequestLocalization(this IApplicationBuilde
...
+            }
+
+            if (cultures == null)
+            {
+                throw new ArgumentNullException(nameof(cultures));
+            }
+
+            if (cultures.Length == 0)
+            {
+                throw new ArgumentException($"The {cultures} cannot be null.", nameof(cultures));
```
hishamco: I 'll move this to the `Resources.resx` at the end

src/Microsoft.AspNetCore.Localization/ApplicationBuilderExtensions.cs
ryanbrandenburg: I don't know the history behind why things are the way the currently are. @DamianEdwards might. I can get on-board with with putting this in the middle-ware.

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
mkArtakMSFT: Do you plan to add tests?

mkArtakMSFT:
@hishamco, we plan to take this into our coming release. It would be great if you could address @rynowak's feedback, so we can move forward with this PR. Thanks!

hishamco:
@mkArtakMSFT sure I'm waiting for @rynowak to clarification ⌚️

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
hishamco: Of course, this will include the case that you mentioned as well as a few other tests

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
rynowak: Ok

src/Microsoft.AspNetCore.Localization/ApplicationBuilderExtensions.cs
rynowak: This isn't the right text. It should say something like "Please provide at least one culture."

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
rynowak: hmm. I looked again at the issue, and I couldn't find it. Ignore this comment I guess.

src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs
hishamco: @mkArtakMSFT [this](https://github.com/aspnet/Localization/pull/369/commits/db4cc145224f4415f3e1de03d9b993b96ef90888#diff-90eeb8c3589c85f68d665deef189a6e4R26) is a proven test for the case that you mentioned above

hishamco:
@rynowak hope this is fine for you after adding a unit & functional tests

hishamco:
Any updates on this?

mkArtakMSFT:
Thanks @hishamco for your effort!

PR #373, opened 2017-05-23
React to StringSegment header parsers

Tratcher:
https://github.com/aspnet/HttpAbstractions/pull/844


ryanbrandenburg:
Is all this being done for performance via `StringSegment`? These don't seem like API's average users would touch, but apparently you'll need more changes as per [this build failure](https://travis-ci.org/aspnet/Localization/jobs/235333566#L550).

Tratcher:
@ryanbrandenburg cross repo breaking change, this change depends on aspnet/HttpAbstractions#844 and travis can't handle that.

PR #374, opened 2017-05-24
Target .NET Standard 2.0 and add .NET Framework 4.6.1 testing

dougbu:


dougbu:
FYI this branch passes locally but will fail on AppVeyor and Travis until UniverseCoherence passes.

build/dependencies.props
```diff
@@ -6,6 +6,7 @@
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
```
dougbu: File contains a `$(CoreFxVersion)` setting but that shouldn't change in this repo. The property is only used in Microsoft.Extensions.* projects targeting .NET Standard 1.3 or earlier.

dougbu:
9b017ad

PR #377, opened 2017-06-02
Remove TaskCache.DefaultCompletedTask

hishamco:


hishamco:
BTW @pranavkm is this fine to go?

khellang:
Doesn't this PR remove the whole concept of a cached `Task<T>`? I'd agree on `Task.CompletedTask`, but not its `Task<T>` relatives.

davidfowl:
Yea, I'm not sure what the intent of this change was.

hishamco:
@davidfowl why it's required since we target Net Standard 2.0?

khellang:
@hishamco Because the two solutions are not functionally equivalent. The existing code caches an instance of `Task<T>`, while your proposed solution [allocates a new `Task<T>`](https://github.com/dotnet/coreclr/blob/906f60ce3c55ab0cc41765d9eeea542ec65ff737/src/mscorlib/src/System/Threading/Tasks/Task.cs#L5054) every time. That's the whole reason for using those APIs in the first place :wink:

hishamco:
You're right, his because the `static`
If I add the `static` again I'm not sure how other repos will shared the same code, it should be in a common namespace

hishamco:
I think we need to remove the [TaskCache](https://github.com/aspnet/Mvc/blob/18785dbed6ef63990f1732c509a19cf8ef533cb5/src/Microsoft.AspNetCore.Mvc.Core/Internal/TaskCache.cs) and [TaskCache<T>](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/Internal/TaskCacheOfT.cs) from MVC only, so I think I should close this now

PR #381, opened 2017-06-08
Remove usage of TaskCache

pranavkm:


src/Microsoft.AspNetCore.Localization/RequestCultureProvider.cs
```diff
@@ -13,6 +13,12 @@ namespace Microsoft.AspNetCore.Localization
     public abstract class RequestCultureProvider : IRequestCultureProvider
     {
         /// <summary>
+        /// Result that indicates that this instance of <see cref="RequestCultureProvider" /> could not determine the
+        /// request culture.
+        /// </summary>
+        protected static readonly Task<ProviderCultureResult> NullResult = Task.FromResult(default(ProviderCultureResult));
```
pranavkm: This removes the last use of `Task<T>` from our code base (and we can remove the package). Given the pattern requires returning this, it made sense to put the field on the base type. I can alternatively make it an `internal` field or put it in an `.Internal` type.

Eilon: Hmm this is probably fine. Can't do any harm really. Or maybe it should have a more specific name, though? Is `NullProviderCultureResult` too long? I think it reads better in the code, though.

pranavkm:
> Is NullProviderCultureResult too long? 

Nah, seems fine given it's narrow use. 🆙 📅

hishamco:
I already made similar PR #377, but I confused little bit, because we need to do similar things in all the repos that uses the `TaskCache`

pranavkm:
@hishamco you want to cache the `Task<ProviderCultureResult>` somewhere which was missing in your PR.

hishamco:
Yep, I planned to make it static and store somewhere else, anyhow your PR looks fine 😄

PR #264, opened 2016-07-06
Get all string experiment

ryanbrandenburg:
@rynowak here is my experiment in abstracting the management of resource strings properly.

I was wondering if maybe the interface should actually be for the strings instead of a stream since what we'd like to have as a final fix for this is a corefx API which would give us all the strings for a culture, rather than a stream.

PR #384, opened 2017-06-14
Permanent fix for the issue #257

hishamco:
The PR #264 was a temporary fix for the issue #257, I create this PR to be a permanent fix for that previous issue

/cc @ryanbrandenburg @pranavkm


hishamco:
I had to add a new property `Count` to `IResourceNamesCache` to make enumerator caches tests passing, I know this is a **breaking change** but I don't find a way to make all the tests passed without adding this property. Let me know if there's another a better way for doing this

hishamco:
@pranavkm could you please have a look to this too, i'm not sure if there's a breaking should go in 2.0.0 or we need to wait until 3.0.0

pranavkm:
@rynowak?

src/Microsoft.Extensions.Localization/Internal/AssemblyResourceStringProvider.cs
```diff
@@ -1,115 +0,0 @@
...
-using System.Collections;
-using System.Collections.Generic;
-using System.Globalization;
-using System.IO;
-using System.Reflection;
-using System.Resources;
-
-namespace Microsoft.Extensions.Localization.Internal
-{
-    public class AssemblyResourceStringProvider : IResourceStringProvider
```
hishamco: I didn't get you?!!

natemcmaster:
As you may already know, we're going to archive this repo. I'm working on moving the code today. Once complete, the Microsoft.Extensions.\* projects will be in https://github.com/aspnet/Extensions, and Microsoft.AspNetCore.\* will be in https://github.com/aspnet/AspNetCore. If this is something you'd like to get merged, please open a new PR against the new repos.

Thanks,
Nate

Issue #372, opened 2017-05-22
Consider to use logging in the Request Culture Providers


hishamco:
Based on the issue #370, I think it would be nice if we can add some logging into the default `RequestCultureProvider`s, specifically when the `CookieRequestCultureProvider` failed to parse the cookie, `QueryRequestCultureProvider` failed to parse the querystring

/cc @ryanbrandenburg @dougbu

Eilon:
@ryanbrandenburg assigning this to you because I've assigned @hishamco 's PR at https://github.com/aspnet/Localization/pull/385 to you to review & merge.

PR #385, opened 2017-06-14
Log warning whenever the cultures are invalid in Request Culture Providers

hishamco:
Address issue #372

### DIFF
--- a/src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
+++ b/src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
@@ -2,9 +2,14 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
 using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Internal;
+using Microsoft.AspNetCore.Localization.Internal;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Localization
 {
@@ -16,6 +21,7 @@ namespace Microsoft.AspNetCore.Localization
         private static readonly char[] _cookieSeparator = new[] { '|' };
         private static readonly string _culturePrefix = "c=";
         private static readonly string _uiCulturePrefix = "uic=";
+        private ILogger _logger;
 
         /// <summary>
         /// Represent the default cookie name used to track the user's preferred culture information, which is ".AspNetCore.Culture".
@@ -45,6 +51,27 @@ namespace Microsoft.AspNetCore.Localization
 
             var providerResultCulture = ParseCookieValue(cookie);
 
+            if (providerResultCulture != null)
+            {
+                Debug.Assert(providerResultCulture.Cultures.Count == 1);
+                Debug.Assert(providerResultCulture.UICultures.Count == 1);
+                
+                var culture = providerResultCulture.Cultures[0].ToString();
+                var uiCulture = providerResultCulture.UICultures[0].ToString();
+                
+                if (!IsValidCulture(culture))
+                {
+                    _logger = _logger ?? httpContext.RequestServices.GetService<ILogger<CookieRequestCultureProvider>>();
+                    _logger?.InvalidCultureName(nameof(CookieRequestCultureProvider), culture);
+                }
+                
+                if (!IsValidCulture(uiCulture))
+                {
+                    _logger = _logger ?? httpContext.RequestServices.GetService<ILogger<CookieRequestCultureProvider>>();
+                    _logger?.InvalidUICultureName(nameof(CookieRequestCultureProvider), uiCulture);
+                }
+            }
+                     
             return Task.FromResult(providerResultCulture);
         }
 
@@ -118,5 +145,8 @@ namespace Microsoft.AspNetCore.Localization
 
             return new ProviderCultureResult(cultureName, uiCultureName);
         }
+
+        private static bool IsValidCulture(string cultureName) =>
+            CultureInfo.GetCultures(CultureTypes.SpecificCultures).Any(c => c.Name.Equals(cultureName));
     }
 }
--- a/src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Localization.Internal
+{
+    internal static class RequestCultureProviderLoggerExtensions
+    {
+        private static readonly Action<ILogger, string, string, Exception> _invalidCulture;
+        private static readonly Action<ILogger, string, string, Exception> _invalidUICulture;
+
+        static RequestCultureProviderLoggerExtensions()
+        {
+            _invalidCulture = LoggerMessage.Define<string, string>(
+                LogLevel.Warning,
+                1,
+                "{requestCultureProvider} contains invalid culture '{cultureName}'.");
+            _invalidUICulture = LoggerMessage.Define<string, string>(
+                LogLevel.Warning,
+                1,
+                "{requestCultureProvider} contains invalid ui-culture '{cultureName}'.");
+        }
+
+        public static void InvalidCultureName(this ILogger logger, string requestCultureProvider, string cultureName)
+        {
+            _invalidCulture(logger, requestCultureProvider, cultureName, null);
+        }
+
+        public static void InvalidUICultureName(this ILogger logger, string requestCultureProvider, string cultureName)
+        {
+            _invalidUICulture(logger, requestCultureProvider, cultureName, null);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Localization/Microsoft.AspNetCore.Localization.csproj
+++ b/src/Microsoft.AspNetCore.Localization/Microsoft.AspNetCore.Localization.csproj
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
 
--- a/test/Microsoft.AspNetCore.Localization.Tests/CookieRequestCultureProviderTest.cs
+++ b/test/Microsoft.AspNetCore.Localization.Tests/CookieRequestCultureProviderTest.cs
@@ -8,6 +8,9 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Localization;
 using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Net.Http.Headers;
 using Xunit;
 
@@ -33,8 +36,10 @@ namespace Microsoft.Extensions.Localization.Tests
                             new CultureInfo("ar-SA")
                         }
                     };
-                    var provider = new CookieRequestCultureProvider();
-                    provider.CookieName = "Preferences";
+                    var provider = new CookieRequestCultureProvider
+                    {
+                        CookieName = "Preferences"
+                    };
                     options.RequestCultureProviders.Insert(0, provider);
 
                     app.UseRequestLocalization(options);
@@ -77,8 +82,10 @@ namespace Microsoft.Extensions.Localization.Tests
                             new CultureInfo("ar-SA")
                         }
                     };
-                    var provider = new CookieRequestCultureProvider();
-                    provider.CookieName = "Preferences";
+                    var provider = new CookieRequestCultureProvider
+                    {
+                        CookieName = "Preferences"
+                    };
                     options.RequestCultureProviders.Insert(0, provider);
                     app.UseRequestLocalization(options);
                     app.Run(context =>
@@ -117,8 +124,10 @@ namespace Microsoft.Extensions.Localization.Tests
                             new CultureInfo("ar-SA")
                         }
                     };
-                    var provider = new CookieRequestCultureProvider();
-                    provider.CookieName = "Preferences";
+                    var provider = new CookieRequestCultureProvider
+                    {
+                        CookieName = "Preferences"
+                    };
                     options.RequestCultureProviders.Insert(0, provider);
                     app.UseRequestLocalization(options);
                     app.Run(context =>
@@ -135,6 +144,65 @@ namespace Microsoft.Extensions.Localization.Tests
                 var client = server.CreateClient();
                 var response = await client.GetAsync(string.Empty);
             }
+        }
+
+        [Theory]
+        [InlineData("??", "en-US", "CookieRequestCultureProvider contains invalid culture '??'.")]
+        [InlineData("en-US", "??", "CookieRequestCultureProvider contains invalid ui-culture '??'.")]
+        public async Task ParseInvalidCookieShouldLogWarning(string culture, string uiCulture, string expectedMessage)
+        {
+            var sink = new TestSink(
+                TestSink.EnableWithTypeName<CookieRequestCultureProvider>,
+                TestSink.EnableWithTypeName<CookieRequestCultureProvider>);
+            var loggerFactory = new TestLoggerFactory(sink, enabled: true);
+            var builder = new WebHostBuilder()
+                .Configure(app =>
+                {
+                    var options = new RequestLocalizationOptions
+                    {
+                        DefaultRequestCulture = new RequestCulture("en-US"),
+                        SupportedCultures = new List<CultureInfo>
+                        {
+                            new CultureInfo("ar-YE")
+                        },
+                        SupportedUICultures = new List<CultureInfo>
+                        {
+                            new CultureInfo("ar-YE")
+                        }
+                    };
+                    var provider = new CookieRequestCultureProvider
+                    {
+                        CookieName = "Preferences"
+                    };
+                    options.RequestCultureProviders.Insert(0, provider);
+                    app.UseRequestLocalization(options);
+                    app.Run(context =>
+                    {
+                        var requestCultureFeature = context.Features.Get<IRequestCultureFeature>();
+                        var requestCulture = requestCultureFeature.RequestCulture;
+                        Assert.Equal("en-US", requestCulture.Culture.Name);
+                        return Task.FromResult(0);
+                    });
+                })
+                .ConfigureServices(services =>
+                {
+                    services.AddSingleton(typeof(ILoggerFactory), loggerFactory);
+                    services.AddSingleton(typeof(ILogger), loggerFactory.CreateLogger<CookieRequestCultureProvider>());
+                });
+
+            using (var server = new TestServer(builder))
+            {
+                var client = server.CreateClient();
+                client.DefaultRequestHeaders.Add("Cookie", new CookieHeaderValue("Preferences", $"c={culture}|uic={uiCulture}").ToString());
+
+                var response = await client.GetAsync(string.Empty);
+            }
+
+            var logMessages = sink.Writes;
+
+            Assert.Equal(1, logMessages.Count);
+            Assert.Equal(LogLevel.Warning, logMessages[0].LogLevel);
+            Assert.Equal(expectedMessage, logMessages[0].State.ToString());
         }
     }
 }
--- a/test/Microsoft.AspNetCore.Localization.Tests/Microsoft.AspNetCore.Localization.Tests.csproj
+++ b/test/Microsoft.AspNetCore.Localization.Tests/Microsoft.AspNetCore.Localization.Tests.csproj
@@ -13,7 +13,9 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Testing" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(TestSdkVersion)" />
     <PackageReference Include="xunit" Version="$(XunitVersion)" />
     <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitVersion)" />
--- a/src/Microsoft.AspNetCore.Localization/QueryStringRequestCultureProvider.cs
+++ b/src/Microsoft.AspNetCore.Localization/QueryStringRequestCultureProvider.cs
@@ -2,9 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
 using System;
+using System.Globalization;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Internal;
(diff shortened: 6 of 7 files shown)


hishamco:
@Eilon is it enough to include the `CookieCultureProvider` only or shall I add the `QueryStringCultureProvider`?

The current implementation should support both of them, but `RouteDataCultureProvider` not included because it has it's own assembly

I will finish this by adding some tests, let me know if there's any further comments

hishamco:
This is strange little bit I tried many times in my laptop the test is passed, now why it pass in AppVeyor and not in Travis?!!

@ryanbrandenburg any idea

hishamco:
@ryanbrandenburg @pranavkm could you please review this, and let me why it partially fails, is it fixed after the commit https://github.com/aspnet/Localization/commit/f2b1e355b397d2a33431f8727e95ba492e59f636?!!

pranavkm:
@hishamco we're close to completion on our 2.0.0 work. The work item's been triaged for the next milestone. I'll have a look at it and merge it in once our dev branch is open for 2.1.0 work.

hishamco:
Thanks

test/Microsoft.AspNetCore.Localization.Tests/CookieRequestCultureProviderTest.cs
```diff
@@ -136,5 +145,62 @@ public class CookieRequestCultureProviderTest
...
+                .ConfigureServices(services =>
+                {
+                    services.AddSingleton(typeof(ILoggerFactory), loggerFactory);
+                    services.AddSingleton(typeof(ILogger), loggerFactory.CreateLogger<CookieRequestCultureProvider>());
+                });
+
+            using (var server = new TestServer(builder))
+            {
+                var client = server.CreateClient();
+                client.DefaultRequestHeaders.Add("Cookie", new CookieHeaderValue("Preferences", "c=??|uic=en-US").ToString());
```
Eilon: Is the `??` in there the thing that makes it invalid? If so, I think it would be good to add a comment before this line to emphasize it.

test/Microsoft.AspNetCore.Localization.Tests/Microsoft.AspNetCore.Localization.Tests.csproj
```diff
@@ -11,9 +11,11 @@
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Localization\Microsoft.AspNetCore.Localization.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup>
```
Eilon: Tailing whitespace 😄

Eilon:
@ryanbrandenburg can you also take a look?

test/Microsoft.AspNetCore.Localization.Tests/CookieRequestCultureProviderTest.cs
hishamco: > Is the ?? in there the thing that makes it invalid?

Yep

We can a comment, but I expect the test name `ParseInvalidCookieShouldLogWarning` is more obvious instead

test/Microsoft.AspNetCore.Localization.Tests/Microsoft.AspNetCore.Localization.Tests.csproj
hishamco: Oops, nice catch

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
```diff
@@ -98,6 +106,16 @@ public static ProviderCultureResult ParseCookieValue(string value)
             var cultureName = potentialCultureName.Substring(_culturePrefix.Length);
             var uiCultureName = potentialUICultureName.Substring(_uiCulturePrefix.Length);
 
+            if (!IsValidCulture(cultureName))
```
pranavkm: Should this go after the null check on Line 101 \ 119? Presumably `GetCultureInfo` will throw a arg null if it's null.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
```diff
@@ -16,6 +21,7 @@ public class CookieRequestCultureProvider : RequestCultureProvider
         private static readonly char[] _cookieSeparator = new[] { '|' };
         private static readonly string _culturePrefix = "c=";
         private static readonly string _uiCulturePrefix = "uic=";
+        private static ILogger _logger;
```
pranavkm: No static fields.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
```diff
@@ -36,6 +42,8 @@ public override Task<ProviderCultureResult> DetermineProviderCultureResult(HttpC
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
+            _logger = httpContext.RequestServices.GetService<ILogger>() ?? NullLogger.Instance;
```
pranavkm: Can we refactor the code to pass this in via the ctor? We could change RequestLocalizationOptions.cs to be initialized via  an options setup that has access to the service container and have it call the right ctor.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: We could make the check in `IsValidCulture()` before
```csharp
CultureInfo.GetCultureInfo(cultureName);
```

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: @pranavkm at the first glance I was thinking to inject the `ILogger` into the constructor, but while the `ParseCookieValue` is static I come up with this idea

hishamco: If I understand you correctly there should be two constructors, Am I right? If yes .. why we can make it one while the `CookieRequestCultureProvider` new up inside the `RequestLocalizationOptions`?

pranavkm: Yeah something like that. We'll leave it no-op the log \ log to NullLogger in the scenario the old ctor is called.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
pranavkm: And treat it as a valid culture? Sure, that sort of seems acceptable.

test/Microsoft.AspNetCore.Localization.Tests/CookieRequestCultureProviderTest.cs
Eilon: The test name definitely helps, but it isn't obvious at first where in the code the "invalid cookie" is - or why it's invalid.

src/Microsoft.AspNetCore.Localization/Microsoft.AspNetCore.Localization.csproj
```diff
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="$(AspNetCoreVersion)" />
```
pranavkm: Is this for the `NullLogger` instance? I'd rather we avoid adding package references and just check for null `ILogger` instances in the code.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
Eilon: BTW I want to make sure there are no breaking changes here, so be careful about adding a ctor if there wasn't one before. If we add a new parameterized ctor, we'd need to add a new empty ctor as well for compat (and make sure that things actually work with both ctors).

src/Microsoft.AspNetCore.Localization/Microsoft.AspNetCore.Localization.csproj
Eilon: Yeah you could just use `_logger?.FooLog(...)` - we do that in tons of places.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
```diff
@@ -2,9 +2,14 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
 
 using System;
+using System.Globalization;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Localization.Internal;
```
Eilon: Ick. What's used from this namespace? We do our best to not reference pseudo-internal types from other assemblies.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
pranavkm: Yup, that's sort of what I was asking @hishamco to do. We'd have a no-arg ctor for back-compat and one which accepts an `ILogger`. We'll call the new ctor via a new options setup type. The breaking change here would be that https://github.com/aspnet/Localization/blob/f650de8cdf5214bcbb2e88210baac0f99ce3d053/src/Microsoft.AspNetCore.Localization/RequestLocalizationOptions.cs#L24 wouldn't be initialized in the ctor, but from options setup. However if you go via the expected code path of pulling options from the DI container, everything should be hunky dory.

test/Microsoft.AspNetCore.Localization.Tests/CookieRequestCultureProviderTest.cs
ryanbrandenburg: Maybe also use a culture shaped string like "no-OP".

Eilon: `no-OP` is probably a valid culture on Win10. I like using a super-obviously-not-a-real-culture for this.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: Another reason for use the current implementation is to avoid breaking change as @Eilon suggest

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: Nope it would be invalid one

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: I use it for the `ParsedCulture()` extension method, we already have something similar [here](https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizer.cs#L10)  for `SearchedLocation()`

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: How we use non `static` field in `ParseCookieValue()`?!!

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
pranavkm: In that case, let's do it after we do all the null-checks. I don't want us to log things if we ultimately wouldn't do anything with it.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
pranavkm: Sorry, I just had a look at the type in entirety. We'd have to pass the logger explicitly in to a separate overload. Essentially we can have it log when `DetermineProviderCultureResult` is invoked using the new ctor.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
pranavkm: Adding the two ctors wouldn't be a breaking change.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: In this case we need to duplicate the code to get the `culture` and `ui-culture` like what we have in `ParseCookieValue()` to use them in logging

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: @pranavkm while the `CookieRequestCultureProvider` needs two constructors, we need somehow to get the `HttpContext` to access the `ILogger` from the DI container
```csharp
public RequestLocalizationOptions(IHttpContextAccessor httpContextAccessor)
{
    var logger = (ILogger)httpContextAccessor.HttpContext
        .RequestServices.GetService(typeof(ILogger));

    RequestCultureProviders = new List<IRequestCultureProvider>
    {
        new QueryStringRequestCultureProvider { Options = this },
        (logger == null ?
            new CookieRequestCultureProvider { Options = this } :
            new CookieRequestCultureProvider(logger) { Options = this }),
        new AcceptLanguageHeaderRequestCultureProvider { Options = this }
    };
}
```
If I made something like the above code snippet, there will be a breaking change, which is not preferred, could you please show me a code snippet for the things in your mind.

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Localization.Internal
```
Eilon: This is in the wrong namespace. The namespace (in general) needs to be based on the assembly's main namespace. Maybe it should be `Microsoft.AspNetCore.Localization.Internal`?

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
```diff
@@ -98,6 +106,16 @@ public static ProviderCultureResult ParseCookieValue(string value)
...
             var uiCultureName = potentialUICultureName.Substring(_uiCulturePrefix.Length);
 
+            if (!IsValidCulture(cultureName))
+            {
+                _logger.ParsedCulture(nameof(CookieRequestCultureProvider), cultureName);
+            }
+
+            if (!IsValidCulture(uiCultureName))
+            {
+                _logger.ParsedCulture(nameof(CookieRequestCultureProvider), uiCultureName);
```
Eilon: These two logs are identical in terms of the message - perhaps they should have some distinction to indicate which culture was being parsed, i.e. Culture vs. UICulture?

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            _parsedCulture = LoggerMessage.Define<string, string>(
+                LogLevel.Warning,
+                1,
+                $"{{requestCultureProvider}} contains invalid culture '{{cultureName}}'.");
+        }
+
+        public static void ParsedCulture(this ILogger logger, string requestCultureProvider, string cultureName)
+        {
+            _parsedCulture(logger, requestCultureProvider, cultureName, null);
```
Eilon: What event id / category name gets used here? I'm having a hard time figuring out what gets emitted in the log when this log is written.

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
hishamco: This is similar to https://github.com/aspnet/Localization/blob/dev/src/Microsoft.Extensions.Localization/Internal/ResourceManagerStringLocalizerLoggerExtensions.cs#L8

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
hishamco: @Eilon the extension is pretty much smiliar to `ResourceManagerStringLocalizerLoggerExtensions`

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: You're right this totally useful

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: @pranavkm could you have a look to this
⌚️

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
Eilon: It's similar, but it's in a different assembly/namespace.

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
Eilon: Hmm ok, but it still isn't clear to me what logger category this goes in?

hishamco: No no .. you are right about this 😄

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
hishamco: You are right about this, it's my fault 😄

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
```diff
@@ -4,23 +4,33 @@
...
+namespace Microsoft.AspNetCore.Localization.Internal
 {
     internal static class RequestCultureProviderLoggerExtensions
     {
         private static readonly Action<ILogger, string, string, Exception> _parsedCulture;
+        private static readonly Action<ILogger, string, string, Exception> _parsedUICulture;
 
         static RequestCultureProviderLoggerExtensions()
         {
             _parsedCulture = LoggerMessage.Define<string, string>(
```
pranavkm: Could you rename these field names too?

pranavkm:
Doesn't look like `CultureInfo.GetCultureInfo` throws on xplat. I don't think we can use this that to determine if a culture is valid.

hishamco:
@pranavkm I tried two different ways with no luck, but finally the test is passed with my third attempt 😄

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        private static readonly Action<ILogger, string, string, Exception> _invalidCulture;
+        private static readonly Action<ILogger, string, string, Exception> _invalidUICulture;
+
+        static RequestCultureProviderLoggerExtensions()
+        {
+            _invalidCulture = LoggerMessage.Define<string, string>(
+                LogLevel.Warning,
+                1,
+                $"{{requestCultureProvider}} contains invalid culture '{{cultureName}}'.");
```
Eilon: If you remove the `$` from these, there's no need to escape the `{` and `}`'s.

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
```diff
@@ -118,5 +145,8 @@ public static ProviderCultureResult ParseCookieValue(string value)
 
             return new ProviderCultureResult(cultureName, uiCultureName);
         }
+
+        private static bool IsValidCulture(string cultureName) =>
+            CultureInfo.GetCultures(CultureTypes.SpecificCultures).Any(c => c.Name == cultureName);
```
Eilon: I think this check needs to be an *identical* check to whatever prompted this change in the first place. That is, when there's an invalid culture, what call used to fail? Let's just do that call, no?

hishamco: May be we can check against the supported cultures, and I think the check that I made is fine

src/Microsoft.AspNetCore.Localization/Internal/RequestCultureProviderLoggerExtensions.cs
hishamco: Why do we need to delete it?!!

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
hishamco: > I think this check needs to be an identical check

@Eilon did you mean shall we do a string casing check?!!

pranavkm: Here's what we do to actually parse the culture (https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs#L156-L173). We could do the same check here to ascertain if the culture if something we'd eventually use. IMO, we could just move the test to the actual middleware and have it complain if we returned a `ProviderCultureResult` which we would not use.

Eilon: Yeah I think what @pranavkm is saying is what I had in mind. I just want to make sure we have only one way of validating/using cultures.

hishamco: Believe it or not I was thinking about that yesterday, so that's why I wrote my comment [above](https://github.com/aspnet/Localization/pull/385#discussion_r134286514)

I will fix it now ..

hishamco: > IMO, we could just move the test to the actual middleware and have it complain if we returned a `ProviderCultureResult` which we would not use.

@pranavkm If I move the check to the `RequestLocalizationMiddleware` I have two points to discuss: 
1. `_options` field need to be `static` - which is the thing that you don't like 😄 - to access the `SupportedCultures`
2. `GetCultureInfo()` accept the supported cultures, in our case we need to check both of `SupportedCultures` & `SupportedUICultures`

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
```diff
@@ -118,5 +145,8 @@ public static ProviderCultureResult ParseCookieValue(string value)
 
             return new ProviderCultureResult(cultureName, uiCultureName);
         }
+
+        private static bool IsValidCulture(string cultureName) =>
+            CultureInfo.GetCultures(CultureTypes.SpecificCultures).Any(c => c.Name.Equals(cultureName));
```
ryanbrandenburg: @Eilon and @pranavkm are we happy with these checks now?

pranavkm: Nope. We have a separate discussion about this here: https://github.com/aspnet/Localization/pull/385#discussion_r134241089.

hishamco: Please have a look to my comment [here](https://github.com/aspnet/Localization/pull/385#discussion_r134309983)

src/Microsoft.AspNetCore.Localization/CookieRequestCultureProvider.cs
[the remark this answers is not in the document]
pranavkm: Isn't this all you need to do: https://github.com/aspnet/Localization/compare/prkrishn/loc-warn?expand=1?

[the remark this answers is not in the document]
hishamco: Seems my understanding was wrong, I though you want to use `GetCultureInfo()` in a new function, anyhow either you accept your PR or I will modify this to include your latest changes alongside with the unit test that I already created

pranavkm:
https://github.com/aspnet/Localization/commit/41ae20525f4aa22490ffb435db2c86e2906a39b0 was merged to address the issue. Thanks for your PR @hishamco

hishamco:
@pranavkm  Can I push my unit tests?

pranavkm:
Sure, could you redo them on top of dev?

PR #393, opened 2017-07-06
Add breakingchanges

jbagga:
Addresses #350 

cc @Eilon @dougbu @ryanbrandenburg


src/Microsoft.Extensions.Localization/breakingchanges.net45.json
```diff
@@ -0,0 +1,22 @@
+[
```
pranavkm: Shouldn't need a net45.json. We aren't cross-compiling this project.

jbagga:
Removed net45 json file

PR #409, opened 2017-08-28
Log warnings when cultures specified by Request Culture Providers are…

### LINKED ISSUE #372: Consider to use logging in the Request Culture Providers

pranavkm:
… unsupported

Fixes #372

### DIFF
--- a/src/Microsoft.AspNetCore.Localization/RequestCultureProviderLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Localization/RequestCultureProviderLoggerExtensions.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information. 
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Localization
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Localization/RequestCultureProviderLoggerExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+                LogLevel.Warning,
+                1,
+                "{requestCultureProvider} returned the following unsupported cultures '{cultures}'.");
+            _unsupportedUICulture = LoggerMessage.Define<string, IList<StringSegment>>(
+                LogLevel.Warning,
+                2,
+                "{requestCultureProvider} returned the following unsupported cultures '{cultures}'.");
+        }
+
+        public static void UnsupportedCulture(this ILogger logger, string requestCultureProvider, IList<StringSegment> cultures)
```
ryanbrandenburg: Should be `UnsupportedCultures` if it takes a list, no?

pranavkm: Sure

Issue #273, opened 2016-08-05
zh-TW should fallback to zh-Hant before zh


luciusli:
zh-CN should also fallback to zh-Hans.

Eilon:
I have a feeling that the logic is quite simple: It just chops off the last segment of the locale string after the `-` character and uses that. I don't think any part of the system actually knows anything about which languages are "related" to other languages.

@ryanbrandenburg do you know any more about how this fallback logic works?

hishamco:
As @Eilon said zh-TW & zh-Hant should fallback to zh, I never seen a framework to do such thing, because you need something in advance (relation between the languages),

ryanbrandenburg:
Actually, we use `CultureInfo.Parent` to walk down the hierarchy rather than string manipulation.

For zh-TW in net451 (there are some differences between frameworks) the hierarchy goes "zh-TW->zh-CHT->zh-Hant->zh" in netcoreapp1.0 it's the same, but omit "zh-CHT". Either framework should reach zh-Hant before zh. The most likely cause of skipping it would be that zh-TW or zh-Hant are not among `SupportedCultures` or `SupportedUICultures`. If zh-TW and zh-Hant are both in your `SupportedCultures` and `SupportedUICultures` and you are still encountering this problem please upload a reproduction app as a GitHub repository so that I can diagnose the problem.

Eilon:
This seems like a complete explanation. Assuming the steps in @ryanbrandenburg 's response are followed, there is no further issue here, so I'm closing the issue.

If the ultimate issue is that the hierarchy of cultures on .NET Core (netcoreapp) is undesirable, please file a new bug at https://github.com/dotnet/corefx/issues/. Thanks!

yufeih:
@ryanbrandenburg 

> Actually, we use CultureInfo.Parent to walk down the hierarchy rather than string manipulation.

It is actually using a simple string manipulation here: https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs#L206

> If zh-TW and zh-Hant are both in your SupportedCultures and SupportedUICultures and you are still encountering this problem please upload a reproduction app as a GitHub repository so that I can diagnose the problem.

I prepared a [PR](https://github.com/aspnet/Localization/pull/416/files) with a test case to reproduce the problem

Issue #415, opened 2017-09-14
FallBackToParentCultures does not respect zh-CN and zh-Hans


yufeih:
When `FallBackToParentCultures` is set, *zh-CN* will not be fallback to *zh-Hans*.

Locale fallback simply trim the last section after dash, this is incorrect:

https://github.com/aspnet/Localization/blob/dev/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs#L206

*zh-Hans* should be a parent locale of *zh-CN*

hishamco:
Could you please refer to issue #273, seems `zh-{cultures}` are more complicated than I knew 😄

aspnet-hello:
This issue was moved to aspnet/Home#2634

PR #416, opened 2017-09-14
Change culture fallback to use corefx parent culture

yufeih:
Uses `CultureInfo.Parent` as fallback parent culture, instead of trimming last part.

Resolves https://github.com/aspnet/Localization/issues/273 and https://github.com/aspnet/Localization/issues/415

@hishamco @ryanbrandenburg

### DIFF
--- a/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Localization/RequestLocalizationMiddleware.cs
@@ -152,13 +152,27 @@ namespace Microsoft.AspNetCore.Localization
             IList<CultureInfo> supportedCultures,
             bool fallbackToParentCultures)
         {
+            CultureInfo cultureInfo;
+
             foreach (var cultureName in cultureNames)
             {
                 // Allow empty string values as they map to InvariantCulture, whereas null culture values will throw in
                 // the CultureInfo ctor
                 if (cultureName != null)
                 {
-                    var cultureInfo = GetCultureInfo(cultureName, supportedCultures, fallbackToParentCultures, currentDepth: 0);
+                    try
+                    {
+                        cultureInfo = CultureInfo.GetCultureInfo(cultureName.ToString());
+                    }
+                    catch (CultureNotFoundException)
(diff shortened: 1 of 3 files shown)


hishamco:
There's a test failed on Travis CI

yufeih:
Right, `CultureInfo.Parent` seems to behave differently on linux

hishamco:
May be, so leaving this PR for now until the `CultureInfo.Parent` is got fixed, @Eilon I think we need to file a bug in Core CLR repo regarding this

yufeih:
`CultureInfo.Parent` on linux is calling [uloc_getParent()](http://transit.iut2.upmf-grenoble.fr/doc/icu-doc/html/uloc_8h.html#ab722eaa3e697c2df034e1a65fba0b06d). `zh-CN` is an alias of `zh-Hans-CN`, `uloc_getParent` does not map aliased names.

hishamco:
@Eilon what's the action needed here, while the implementation of the `CultureInfo.Parent` seems different in Linux

mkArtakMSFT:
@hishamco, is the fact that the CultureInfo.Parent behaves differently on Linux causes the tests to fail? Can the tests be updated to account for it?

hishamco:
I was expected it's a Core CLR bug, so what's the suitable solution in this case, while the tests are not target a specific OS like what we have in the past

hishamco:
@ryanbrandenburg what we can do regarding this PR?

ryanbrandenburg:
Your tests seem to fail on Ubuntu, likely due to aspnet/Localization#273.

hishamco:
@yufeih any updates on this?

hishamco:
@ryanbrandenburg what's the action needed here?

ryanbrandenburg:
The action needed is to correct either the code or the tests as appropriate to make the tests pass, and resolve the merge conflict.

hishamco:
Seems @yufeih not respond from 4 months after I ping him

ryanbrandenburg:
Yeah, I'm going to close this PR as inactive.

PR #435, opened 2018-01-10
Allow setting RootNamespace in ResourceLocationAttribute

ryanbrandenburg:
This fixes the problem of us lacking a RootNamespace at runtime. I've made it a separate attribute so that when we create it using a task in the future people won't have to worry about the task interfeering with their `ResourceLocationAttribute`.

### DIFF
--- a/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
+++ b/src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Extensions.Localization
                 throw new ArgumentNullException(nameof(typeInfo));
             }
 
-            return GetResourcePrefix(typeInfo, new AssemblyName(typeInfo.Assembly.FullName).Name, _resourcesRelativePath);
+            return GetResourcePrefix(typeInfo, GetRootNamespace(typeInfo.Assembly), GetResourcePath(typeInfo.Assembly));
         }
 
         /// <summary>
@@ -94,9 +94,17 @@ namespace Microsoft.Extensions.Localization
                 throw new ArgumentNullException(nameof(baseNamespace));
             }
 
-            return string.IsNullOrEmpty(resourcesRelativePath)
-                ? typeInfo.FullName
-                : baseNamespace + "." + resourcesRelativePath + TrimPrefix(typeInfo.FullName, baseNamespace + ".");
+            if (string.IsNullOrEmpty(resourcesRelativePath))
+            {
+                return typeInfo.FullName;
+            }
+            else
+            {
+                // This expectation is defined by dotnet's automatic resource storage.
+                // We have to conform to "{RootNamespace}.{ResourceLocation}.{FullTypeName - AssemblyName}".
+                var assemblyName = new AssemblyName(typeInfo.Assembly.FullName).Name;
+                return baseNamespace + "." + resourcesRelativePath + TrimPrefix(typeInfo.FullName, assemblyName + ".");
+            }
         }
 
         /// <summary>
@@ -119,8 +127,9 @@ namespace Microsoft.Extensions.Localization
 
             var assemblyName = new AssemblyName(baseNamespace);
             var assembly = Assembly.Load(assemblyName);
+            var rootNamespace = GetRootNamespace(assembly);
             var resourceLocation = GetResourcePath(assembly);
-            var locationPath = baseNamespace + "." + resourceLocation;
+            var locationPath = rootNamespace + "." + resourceLocation;
 
             baseResourceName = locationPath + TrimPrefix(baseResourceName, baseNamespace + ".");
 
@@ -141,11 +150,10 @@ namespace Microsoft.Extensions.Localization
             }
 
             var typeInfo = resourceSource.GetTypeInfo();
-            var assembly = typeInfo.Assembly;
-            var assemblyName = new AssemblyName(assembly.FullName);
-            var resourcePath = GetResourcePath(assembly);
 
-            var baseName = GetResourcePrefix(typeInfo, assemblyName.Name, resourcePath);
+            var baseName = GetResourcePrefix(typeInfo);
+
+            var assembly = typeInfo.Assembly;
 
             return _localizerCache.GetOrAdd(baseName, _ => CreateResourceManagerStringLocalizer(assembly, baseName));
         }
@@ -215,6 +223,23 @@ namespace Microsoft.Extensions.Localization
         protected virtual ResourceLocationAttribute GetResourceLocationAttribute(Assembly assembly)
         {
             return assembly.GetCustomAttribute<ResourceLocationAttribute>();
+        }
+
+        /// <summary>Gets a <see cref="RootNamespaceAttribute"/> from the provided <see cref="Assembly"/>.</summary>
+        /// <param name="assembly">The assembly to get a <see cref="RootNamespaceAttribute"/> from.</param>
+        /// <returns>The <see cref="RootNamespaceAttribute"/> associated with the given <see cref="Assembly"/>.</returns>
+        /// <remarks>This method is protected and virtual for testing purposes only.</remarks>
+        protected virtual RootNamespaceAttribute GetRootNamespaceAttribute(Assembly assembly)
+        {
+            return assembly.GetCustomAttribute<RootNamespaceAttribute>();
+        }
+
+        private string GetRootNamespace(Assembly assembly)
+        {
+            var rootNamespaceAttribute = GetRootNamespaceAttribute(assembly);
+
+            return rootNamespaceAttribute?.RootNamespace ??
+                new AssemblyName(assembly.FullName).Name;
         }
 
         private string GetResourcePath(Assembly assembly)
--- a/src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
+++ b/src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
@@ -1,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved. 
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Provides the RootNamespace of an Assembly. The RootNamespace of the assembly is used by Localization to
+    /// determine the resource name to look for when RootNamespace differs from the AssemblyName.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+    public class RootNamespaceAttribute : Attribute
+    {
+        /// <summary>
+        /// Creates a new <see cref="RootNamespaceAttribute"/>.
+        /// </summary>
+        /// <param name="rootNamespace">The RootNamespace for this Assembly.</param>
+        public RootNamespaceAttribute(string rootNamespace)
+        {
+            if (string.IsNullOrEmpty(rootNamespace))
+            {
+                throw new ArgumentNullException(nameof(rootNamespace));
+            }
+
+            RootNamespace = rootNamespace;
+        }
+
+        /// <summary>
+        /// The RootNamespace of this Assembly. The RootNamespace of the assembly is used by Localization to
+        /// determine the resource name to look for when RootNamespace differs from the AssemblyName.
+        /// </summary>
+        public string RootNamespace { get; }
+    }
+}
--- a/test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
+++ b/test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
@@ -17,21 +17,30 @@ namespace Microsoft.Extensions.Localization.Tests
     {
         private ResourceLocationAttribute _resourceLocationAttribute;
 
+        private RootNamespaceAttribute _rootNamespaceAttribute;
+
         public Assembly Assembly { get; private set; }
         public string BaseName { get; private set; }
 
         public TestResourceManagerStringLocalizerFactory(
             IOptions<LocalizationOptions> localizationOptions,
             ResourceLocationAttribute resourceLocationAttribute,
+            RootNamespaceAttribute rootNamespaceAttribute,
             ILoggerFactory loggerFactory)
             : base(localizationOptions, loggerFactory)
         {
             _resourceLocationAttribute = resourceLocationAttribute;
+            _rootNamespaceAttribute = rootNamespaceAttribute;
         }
 
         protected override ResourceLocationAttribute GetResourceLocationAttribute(Assembly assembly)
         {
             return _resourceLocationAttribute;
+        }
+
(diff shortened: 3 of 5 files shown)


src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System;
+
+namespace Microsoft.Extensions.Localization
+{
+    /// <summary>
+    /// Provides the RootNamespace of an Assembly.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+    public class RootNamespaceAttribute : Attribute
```
hishamco: I'm not sure if this is a correct place for such property!!

hishamco:
IMHO adding this to `LocalizationOptions` will be better
```csharp
services.AddLocalization(options => {
     options.ResourcesPath = "My/Resources";
     options.RootNamespace = "Alternate Namespace";
});
```

ryanbrandenburg:
@hishamco that doesn't solve the problem for Class libraries, and once we get the task which automatically adds the attribute to your assembly at compile time running, it would only be an extra place you have to update the namespace if you change it.

rynowak:
> IMHO adding this to LocalizationOptions will be better

If it were that simple, I'd totally agree with you 😆 

The root namespace isn't global, it's per assembly

ryanbrandenburg:
Because if an assembly already manually creates the `ResourceLocationAttribute` (which you can only have one of) we don't want them to have problems when we start auto-generating the attribute for RootNamespace.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -68,7 +68,7 @@ protected virtual string GetResourcePrefix(TypeInfo typeInfo)
                 throw new ArgumentNullException(nameof(typeInfo));
             }
 
-            return GetResourcePrefix(typeInfo, new AssemblyName(typeInfo.Assembly.FullName).Name, _resourcesRelativePath);
+            return GetResourcePrefix(typeInfo, GetRootNamespace(typeInfo.Assembly), GetResourcePath(typeInfo.Assembly));
```
kichalla: Why the change `GetResourcePath(typeInfo.Assembly)`? This is a change in behavior?

dougbu: Also, why bother changing the `baseNamespace` argument in a call when you're changing the called overload to ignore `baseNamespace`?

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -93,10 +93,11 @@ protected virtual string GetResourcePrefix(TypeInfo typeInfo, string baseNamespa
             {
                 throw new ArgumentNullException(nameof(baseNamespace));
             }
+            var assemblyName = new AssemblyName(typeInfo.Assembly.FullName).Name;
 
             return string.IsNullOrEmpty(resourcesRelativePath)
                 ? typeInfo.FullName
-                : baseNamespace + "." + resourcesRelativePath + TrimPrefix(typeInfo.FullName, baseNamespace + ".");
+                : baseNamespace + "." + resourcesRelativePath + TrimPrefix(typeInfo.FullName, assemblyName + ".");
```
dougbu: Why does this method now ignore `baseNamespace` (except for the `null` or empty check)?

test/ResourcesClassLibraryWithAttribute/ResourcesClassLibraryWithAttribute.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>Alternate.Namespace</RootNamespace>
```
dougbu: Is this change necessary?

`$(RootNamespace)` normally only controls the namespace of generated files such as `Resources.Designer.cs` or new files in the project. Since Localization uses only the resources and not the generated class, what matters here?

src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// Creates a new <see cref="RootNamespaceAttribute"/>.
+        /// </summary>
+        /// <param name="rootNamespace">The RootNamespace for this Assembly.</param>
```
dougbu: This comment doesn't mean much. Describe what the root namespace controls.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -217,6 +218,20 @@ protected virtual ResourceLocationAttribute GetResourceLocationAttribute(Assembl
             return assembly.GetCustomAttribute<ResourceLocationAttribute>();
         }
 
+        protected virtual RootNamespaceAttribute GetRootNamespaceAttribute(Assembly assembly)
```
dougbu: In what scenario would users (not our tests) override this method? Leaning toward `internal virtual` instead of exposing this publicly.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: I spoke with @dougbu in person and we agreed against his comments. @kichalla we changed to use GetResourcePath here because it's correct. Otherwise we would be ignoring the ResourceLocationAttribute in some cases.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: Spoke with @dougbu in person and we now agree this is correct.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: I chose `protected virutal` for the same reasons I did for `GetResourceLocationAttribute` just above this. The compromise we reached then was to include a warning against using it in the remarks, which I'll add in my reaction commit.

test/ResourcesClassLibraryWithAttribute/ResourcesClassLibraryWithAttribute.csproj
ryanbrandenburg: This change is necessary. To prove it remove `[assembly: RootNamespace("Alternate.Namespace")]` from `ResourcesClassLibraryWithAttribute\AssemblyInfo.cs` in this PR and watch resources not be found. This is because dotnet uses RootNamespace as part of the name when automatically adding resources to the assembly, but `RootNamespace` is not available at runtime.

ryanbrandenburg:
🆙:date:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -94,9 +94,17 @@ protected virtual string GetResourcePrefix(TypeInfo typeInfo, string baseNamespa
                 throw new ArgumentNullException(nameof(baseNamespace));
             }
 
-            return string.IsNullOrEmpty(resourcesRelativePath)
-                ? typeInfo.FullName
-                : baseNamespace + "." + resourcesRelativePath + TrimPrefix(typeInfo.FullName, baseNamespace + ".");
+            if (string.IsNullOrEmpty(resourcesRelativePath))
+            {
+                return typeInfo.FullName;
```
kichalla: Should this consider the `baseNamespace` here as one could have a RootNamespaceAttribute but no resources relative path?

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -94,9 +94,17 @@ protected virtual string GetResourcePrefix(TypeInfo typeInfo, string baseNamespa
...
+            }
+            else
+            {
+                // This expectation is defined by dotnet's automatic resource storage.
+                // We have to conform to "{RootNamespace}.{ResourceLocation}.{FullTypeName - AssemblyName}".
+                var assemblyName = new AssemblyName(typeInfo.Assembly.FullName).Name;
+                return baseNamespace + "." + resourcesRelativePath + TrimPrefix(typeInfo.FullName, assemblyName + ".");
```
kichalla: Trim the `baseNamespace`, `resourcesRelativePath` for `.` here and any other methods that users(as they are open for extension) can passing in value.

test/ResourcesClassLibraryWithAttribute/ResourcesClassLibraryWithAttribute.csproj
dougbu: Got it

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: To my knowledge if you don't have a `resourcesRelativePath` the the resource name is just the full type name.

src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Provides the RootNamespace of an Assembly.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+    public class RootNamespaceAttribute : Attribute
+    {
+        /// <summary>
+        /// Creates a new <see cref="RootNamespaceAttribute"/>.
+        /// </summary>
+        /// <param name="rootNamespace">The RootNamespace for this Assembly. The RootNamespace of the assembly is used by Localization to
+        /// determine the resource name to look for when RootNamespace differs from the AssemblyName.</param>
```
dougbu: These comments will be much less visible than those for the class. (One of the general `Attribute` oddities.) Suggest moving the second sentence up into the class' `<summary>`.

src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                throw new ArgumentNullException(nameof(rootNamespace));
+            }
+
+            RootNamespace = rootNamespace;
+        }
+
+        /// <summary>
+        /// The RootNamespace of this Assembly. The RootNamespace of the assembly is used by Localization to
+        /// determine the resource name to look for when RootNamespace differs from the AssemblyName.
```
dougbu: 🥇

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
```diff
@@ -111,6 +122,34 @@ public void Create_FromType_ReturnsNewResultForDifferentType()
             Assert.NotSame(result1, result2);
         }
 
+        [Fact]
+        public void Create_ResourceLocationAttribute_UsesRootNamespace()
```
dougbu: Need a test enshrining the has `RootNamespaceAttribute` / doesn't have `ResourceLocationAttribute` behaviour.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
ryanbrandenburg: Good point that we need to start doing that, but we should do so holistically for all such inputs at once and add tests, so I've created aspnet/Home#2792 since it seems like a larger item.

test/Microsoft.Extensions.Localization.Tests/ResourceManagerStringLocalizerFactoryTest.cs
ryanbrandenburg: I don't know what you mean by this, could you expand?

dougbu: I mean this test illustrates what happens when both attributes are in play. From your comments on the product code, I understand that `[assembly: RootNamespace(...)]` is a no-op when there's no `ResourceLocationAttribute` in play. Need a test demonstrating that.

ryanbrandenburg:
:up::date:

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
kichalla: Why not write a test for this case? or you can find out from a test app in VS

src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
dougbu: Done above. Thanks.

src/Microsoft.Extensions.Localization/RootNamespaceAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+            if (string.IsNullOrEmpty(rootNamespace))
+            {
+                throw new ArgumentNullException(nameof(rootNamespace));
+            }
+
+            RootNamespace = rootNamespace;
+        }
+
+        /// <summary>
+        /// The RootNamespace of this Assembly.
```
dougbu: Suggest leaving the explanation here too.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
dougbu: `Create_ResourceLocationAttribute_RootNamespaceIgnoredWhenNoLocation` has an odd name but is that test.

src/Microsoft.Extensions.Localization/ResourceManagerStringLocalizerFactory.cs
```diff
@@ -217,6 +225,24 @@ protected virtual ResourceLocationAttribute GetResourceLocationAttribute(Assembl
...
+            return assembly.GetCustomAttribute<RootNamespaceAttribute>();
+        }
+
+        private string GetRootNamespace(Assembly assembly)
+        {
+            var rootNamespaceAttribute = GetRootNamespaceAttribute(assembly);
+
+            return rootNamespaceAttribute?.RootNamespace == null
+                ? new AssemblyName(assembly.FullName).Name
+                : rootNamespaceAttribute.RootNamespace;
```
dougbu: Completely up to you: `rootNamespaceAttribute?.RootNamespace ?? new AssemblyName(assembly.FullName).Name`

ryanbrandenburg: 👍

ryanbrandenburg:
@kichalla do you approve?

PR #438, opened 2018-01-26
Sanitize user input for ResourceLocation and RootNamespace attributes

hishamco:
Fixes https://github.com/aspnet/Home/issues/2792

### DIFF
--- a/src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
+++ b/src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
+using System.Linq;
 
 namespace Microsoft.Extensions.Localization
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -22,6 +24,11 @@ public ResourceLocationAttribute(string resourceLocation)
                 throw new ArgumentNullException(nameof(resourceLocation));
             }
 
+            if (resourceLocation.IndexOfAny(InvalidPathChars) > -1)
```
hishamco: I'm not sure why `Path.InvalidPathChars()` is not working as expected here?!!

src/Microsoft.Extensions.Localization/ResourceLocationAttribute.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Extensions.Localization
     [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
     public class ResourceLocationAttribute : Attribute
     {
+        private static readonly char[] InvalidPathChars = { '<', '>', ':', ',', '"', '/', '|', '?', '*', '.', '[', ']', '=' };
```
hishamco: Seems '/' causes an exception in non-windows operating systems

mkArtakMSFT:
Thanks @hishamco. Closing this PR per the comment on the related issue.

PR #442, opened 2018-05-02
[Design] Start using global.json and netcoreapp2.2

ryanbrandenburg:


Directory.Build.targets
```diff
@@ -1,7 +1,11 @@
 ﻿<Project>
   <PropertyGroup>
-    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">$(MicrosoftNETCoreApp20PackageVersion)</RuntimeFrameworkVersion>
-    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">$(MicrosoftNETCoreApp21PackageVersion)</RuntimeFrameworkVersion>
+    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.2' ">$(MicrosoftNETCoreApp22PackageVersion)</RuntimeFrameworkVersion>
     <NETStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard2.0' ">$(NETStandardLibrary20PackageVersion)</NETStandardImplicitPackageVersion>
   </PropertyGroup>
+
+  <!-- Don't police what version of NetCoreApp we use -->
```
natemcmaster: Should we add a follow-up issue to remove this? Might be good to mention that this is a workaround for now.

natemcmaster (approved):
Looks good. Just one comment.

test/Directory.Build.props
```diff
@@ -2,9 +2,9 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <DeveloperBuildTestTfms>netcoreapp2.1</DeveloperBuildTestTfms>
+    <DeveloperBuildTestTfms>netcoreapp2.2</DeveloperBuildTestTfms>
     <StandardTestTfms>$(DeveloperBuildTestTfms)</StandardTestTfms>
-    <StandardTestTfms Condition=" '$(DeveloperBuild)' != 'true' ">netcoreapp2.1;netcoreapp2.0</StandardTestTfms>
+    <StandardTestTfms Condition=" '$(DeveloperBuild)' != 'true' ">netcoreapp2.2</StandardTestTfms>
```
pranavkm: If you'd like to keep this line around (it kinda no-ops now), consider rewriting it as 

`<StandardTestTfms Condition=" '$(DeveloperBuild)' != 'true' ">$(StandardTestTfms)</StandardTestTfms>`

build/repo.props
```diff
@@ -12,7 +12,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp20PackageVersion)" />
-    <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp21PackageVersion)" />
+    <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp22PackageVersion)" />
```
pranavkm: @natemcmaster super unrelated - but is it possible to move installing NETFx on xplat as a requirement here? If we remove `net461` from appearing in the sample on xplat, we could make it so that it's only installed on repos that cross-compile sources.

natemcmaster: Would be cool, but not something we can automate AFAIK. Netfx *runtime* installations require admin privileges.

What would be possible, though, is using netfx *reference assemblies* from a package instead of requiring them to be installed on the machine. RepoToolset (dotnet/arcade) already has support for this. I'm hoping to adopt it soon-ish.

pranavkm: I don't see a sudo here - https://github.com/aspnet/BuildTools/blob/dev/files/KoreBuild/KoreBuild.sh#L93-L99?

natemcmaster: This script installs netfx reference assemblies for Linux only. This is what I meant by providing netfx reference assemblies from a nuget package. See https://github.com/aspnet/BuildTools/issues/399 for some more context on this. We decided not to do anything at the moment since there are tentative plans to build this into Microsoft.NET.Sdk. See https://github.com/dotnet/sdk/issues/335 and https://github.com/dotnet/designs/pull/33

.gitignore
```diff
@@ -33,4 +33,3 @@ nuget.exe
 *launchSettings.json
 **/Resources/*.Designer.cs
 .vscode/
-global.json
```
natemcmaster: After looking closer at https://github.com/aspnet/BuildTools/pull/661, I think you should split the netcoreapp2.2 update and global.json into separate pieces of work. Let's do the netcoreapp2.2 work first (I think @pranavkm is working on that already, maybe) and address the design issues on https://github.com/aspnet/BuildTools/pull/661 before we start committing the global.json file to source

natemcmaster (changes requested):

build/dependencies.props
```diff
@@ -3,24 +3,23 @@
...
-    <MicrosoftExtensionsDependencyInjectionAbstractionsPackageVersion>2.2.0-preview1-34066</MicrosoftExtensionsDependencyInjectionAbstractionsPackageVersion>
-    <MicrosoftExtensionsDependencyInjectionPackageVersion>2.2.0-preview1-34066</MicrosoftExtensionsDependencyInjectionPackageVersion>
-    <MicrosoftExtensionsLoggingAbstractionsPackageVersion>2.2.0-preview1-34066</MicrosoftExtensionsLoggingAbstractionsPackageVersion>
-    <MicrosoftExtensionsLoggingConsolePackageVersion>2.2.0-preview1-34066</MicrosoftExtensionsLoggingConsolePackageVersion>
-    <MicrosoftExtensionsLoggingPackageVersion>2.2.0-preview1-34066</MicrosoftExtensionsLoggingPackageVersion>
-    <MicrosoftExtensionsLoggingTestingPackageVersion>2.2.0-preview1-34066</MicrosoftExtensionsLoggingTestingPackageVersion>
-    <MicrosoftExtensionsOptionsPackageVersion>2.2.0-preview1-34066</MicrosoftExtensionsOptionsPackageVersion>
-    <MicrosoftNETCoreApp20PackageVersion>2.0.0</MicrosoftNETCoreApp20PackageVersion>
-    <MicrosoftNETCoreApp21PackageVersion>2.2.0-preview1-26424-04</MicrosoftNETCoreApp21PackageVersion>
+    <InternalAspNetCoreSdkPackageVersion>2.2.0-preview1-t000</InternalAspNetCoreSdkPackageVersion>
```
natemcmaster:
```
error NU1102: Unable to find package Internal.AspNetCore.Sdk with version (>= 2.2.0-preview1-t000) 
```

